PR #76, opened 2018-08-13
Vectorized VLM implementation and removed Fortran; used hwangJT's code as base

johnjasa:


johnjasa:
Merging even though coverage decreased; will fix during workshop

PR #105, opened 2018-08-14
Vectorization and sparsity done

meliani09:
Vectorization and sparsity done for lift_coeff_2D


johnjasa:
The sparse derivatives are not set up correctly; the sizes don't match. Look at `test_aero.py` for example, I'll stop by and help as needed.

You can run the tests locally by calling `testflo .`

johnjasa:
Nice! But it still fails on Python3 due to an incorrect call to `range`: https://travis-ci.org/mdolab/OpenAeroStruct/jobs/416090163

johnjasa:
Thanks guys, this worked out great!

PR #111, opened 2018-08-15
fixed erroneous NANing tests

### LINKED ISSUE #99: Switch to OpenMDAO's asserts for all deriv checks
Numpy's assert_almost_equal doesn't fail with NANs while OM's does so switch to those throughout

johnjasa:
Fixes #99


johnjasa:
Merging despite coverage decrease because we don't need to test deprecated components

PR #113, opened 2018-08-15
updated viewmat and example_deriv_check accordingly

nwu63:


johnjasa:
Merging despite small coverage decrease since this file isn't tested anyway

PR #145, opened 2018-08-17
added writeMesh and getFullMesh to geometry utils

nwu63:


johnjasa:
Merging despite tests seemingly failing due to OM bug

PR #152, opened 2018-08-21
Adding fuel vol constraint and test with example optimization

shamsheersc19:
Also updating wingbox analysis test with wave drag.


johnjasa:
Merging despite tests seemingly failing due to OM bug

PR #155, opened 2018-08-21
Updated Travis to pin OM to 2.4.0 to avoid error during test_multipoi…

johnjasa:
…nt_aero


johnjasa:
Coverage is weird; merging in for now, will try to resolve

Issue #131, opened 2018-08-16
Add functionality for user-specified S_ref_total


nwu63:
For computation of aircraft coefficients

johnjasa:
Fixed with #162

PR #162, opened 2018-08-22
Added option to specify Sref, fixed units and minor edits

nwu63:
Addresses #131.

### DIFF
--- a/openaerostruct/functionals/total_aero_performance.py
+++ b/openaerostruct/functionals/total_aero_performance.py
@@ -14,12 +14,12 @@ class TotalAeroPerformance(Group):
 
     def initialize(self):
         self.options.declare('surfaces', types=list)
+        self.options.declare('user_specified_Sref', types=bool)
 
     def setup(self):
         surfaces = self.options['surfaces']
-        user_supplied_S_ref = False
-        
-        if not user_supplied_S_ref:
+
+        if not self.options['user_specified_Sref']:
             self.add_subsystem('sum_areas',
                 SumAreas(surfaces=surfaces),
                 promotes_inputs=['*S_ref'],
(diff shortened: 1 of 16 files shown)


openaerostruct/functionals/total_aero_performance.py
```diff
@@ -17,11 +17,13 @@ def initialize(self):
 
     def setup(self):
         surfaces = self.options['surfaces']
-
-        self.add_subsystem('sum_areas',
-             SumAreas(surfaces=surfaces),
-             promotes_inputs=['*S_ref'],
-             promotes_outputs=['S_ref_total'])
+        user_supplied_S_ref = False
```
johnjasa: If this is set as False here, then directly used in the logic, it's always the case. Is that intended behavior?

openaerostruct/structures/create_rhs.py
```diff
@@ -39,7 +39,7 @@ def setup(self):
         self.ny = surface['num_y']
 
         self.add_input('total_loads', val=np.zeros((self.ny, 6)), units='N')# dtype=data_type))
-        self.add_input('element_weights', val=np.ones((self.ny-1)), units='N')# dtype=data_type))
+        # self.add_input('element_weights', val=np.ones((self.ny-1)), units='N')# dtype=data_type))
```
johnjasa: Just delete this line here

openaerostruct/functionals/total_aero_performance.py
nwu63: Did you want this as an option which is passed in from the run script? I wasn't sure how to do that without cluttering up the whole thing.

johnjasa: Right now the user would have to modify total_aero_performance.py to change what they're using? If that's the case, then yes, we do want it as an option.

johnjasa:
Merging despite Travis timing out; we need to shorten the testing cycle

Issue #80, opened 2018-08-13
Differentiate von Mises stress calculation


johnjasa:
Translate this part to Python:
https://github.com/mdolab/OpenAeroStruct/blob/master/openaerostruct/structures/vonmises_tube.py#L104

Can probably use Dr. Hwang's version from here as a base: https://github.com/hwangjt/OpenAeroStruct/blob/v2/openaerostruct_v2/structures/components/fea_vonmises_comp.py

JustinSGray:
fixed by #164

PR #164, opened 2018-08-22
cleanup of the tests, and update to OpenMDAO's partials assert. 

JustinSGray:
There was nothing wrong with the derivs in Sandy's derivs as far as I could tell.. but I was able to clean things up a bit and added a few extra derivs in another component. 

Updated to OpenMDAO's partial derivs assert too, because I needed the better error messages. 

In a bunch of structural tests, I turned the stiffness way down... the magnitudes were making the abs error for the derivs really big, even when the relative error was like 1e-9.


JustinSGray:
looking at the change log, now I remember. I messed with the scaling on the objective a bit to make things converge a bit better on that one problematic test. But on the way to fixing it I kind of went down the rabbit hole a bit...

johnjasa:
Merging in even though doc build failed; need to revamp how that's done anyway.

JustinSGray:
sorts out #80

PR #172, opened 2018-08-24
Connecting load_factors to functionals for multipoint aerostruct 

shamsheersc19:
Structural and element weights were computed outside the aerostruct points for multipoint. So the weights of the wing structure did not change with the load factor. Fixing this in the functionals.

+ fixing Breguet range (converting N to kg should not use the load factor)

### DIFF
--- a/openaerostruct/functionals/breguet_range.py
+++ b/openaerostruct/functionals/breguet_range.py
@@ -36,18 +36,19 @@ class BreguetRange(ExplicitComponent):
             name = surface['name']
             self.add_input(name + '_structural_weight', val=1., units='N')
 
-        self.add_input('CT', val=1., units='1/s')
-        self.add_input('CL', val=1.)
-        self.add_input('CD', val=1.)
-        self.add_input('a', val=1., units='m/s')
-        self.add_input('R', val=1., units='m')
-        self.add_input('M', val=1.)
-        self.add_input('W0', val=1., units='kg')
-        self.add_input('load_factor', val=1.)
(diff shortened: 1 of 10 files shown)


shamsheersc19:
V1 OAS + wingbox MP cruise + 2.5g cases match now!!! 

Not sure why the deriv tests are failing.

openaerostruct/functionals/breguet_range.py
```diff
@@ -36,18 +36,19 @@ def setup(self):
...
+        self.add_input('a', val=100., units='m/s')
+        self.add_input('R', val=3000., units='m')
+        self.add_input('M', val=1.2)
+        self.add_input('W0', val=200., units='kg')
+        self.add_input('load_factor', val=1.05)
 
         self.add_output('fuelburn', val=1., units='kg')
 
         self.declare_partials('*', '*')
+        self.set_check_partial_options(wrt='*', method='fd', form='central')
```
johnjasa: Can you get rid of this line so we're not checking against fd? Or did you need that instead of cs?

openaerostruct/functionals/equilibrium.py
```diff
@@ -60,6 +60,7 @@ def setup(self):
         self.declare_partials('total_weight','fuelburn')
         self.declare_partials('total_weight','load_factor')
 
+        self.set_check_partial_options(wrt='*', method='fd', form='central')
```
johnjasa: See if you can get rid of this one as well

PR #193, opened 2018-09-05
Great speed improvement for optimization cases by using csc assembled jacobian

johnjasa:


johnjasa:
Merging in, Coveralls seems to have just bugged out with the previous coverage percentage

Issue #198, opened 2018-09-10
Fix struct-only with wingbox model


nwu63:

johnjasa:
Fixed by #200

PR #200, opened 2018-09-10
added logic for wingbox analysis, added test

nwu63:
Running wingbox analysis with `SpatialBeamAlone` now checks for `distributed_fuel_weight` and `struct_weight_relief`. However currently we still automatically add fuel vol comp for all wingbox cases.


johnjasa:
Fixes #198

PR #227, opened 2018-09-19
Sparsified derivative on geo radius comp.

### LINKED ISSUE #221: Change radius_comp derivs to sparse
The current formulation is unnecessarily using a dense matrix for one declare_partials call:

https://github.com/mdolab/OpenAeroStruct/blob/master/openaerostruct/geometry/radius_comp.py#L33

Kenneth-T-Moore:
closes #221

On a large mesh (177x51) for model with just this component:
- Execution time for 10 calls to linearize:  1.0 seconds -> .001 seconds
- Memory footprint during check partials: 240 MB -> 180 MB.

### DIFF
--- a/openaerostruct/geometry/radius_comp.py
+++ b/openaerostruct/geometry/radius_comp.py
@@ -23,14 +23,21 @@ class RadiusComp(ExplicitComponent):
     def setup(self):
         surface = self.options['surface']
 
-        self.nx, self.ny = surface['mesh'].shape[0],surface['mesh'].shape[1]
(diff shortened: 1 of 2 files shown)


Kenneth-T-Moore:
CI failure looks unrelated to me, but probably deserves an issue.

johnjasa:
Huh, I wonder how it could pass for one python and not the other. I'll check it out.
Nice stats!

johnjasa:
Woops, accidentally hit close instead of merge.

PR #304, opened 2020-03-10
Fix travis

nwu63:
## Purpose
I migrated openAeroStruct from Travis-CI.org to Travis-CI.com today. After that, the build started failing, partly because openMDAO released v3 recently and that broke a bunch of API. I've pinned `openmdao<3` and `testflo==1.3.6` for now. I've also added python 3.7 since that's now available through miniconda.

The tests now pass on travis, although with an openFabrics warning that wasn't there before. Not 100% why that shows up now. Anyhow, this PR should be squashed when merged.

I think the next step would be to actually address those API changes. Then we can finally piggy-back off of openMDAO and ditch py2 also.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ install:
 - export PATH=$HOME/.local/bin:$PATH;
 
 # install openmdao
-- pip install "openmdao[all]<3";
+- pip install "openmdao[dev,doc]<3";
 
 # install openaerostruct itself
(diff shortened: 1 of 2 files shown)


nwu63:
@johnjasa could you take a quick look? I'm not sure if the docs build/deploy stuff is okay with the changes. Is it fine to just do that on the python 3.7 build? And will coveralls be okay too?

Issue #308, opened 2020-04-20
Reynolds number partial derivative possibly incorrect


EricUrbineer:
## Type of issue
Potential bugfix

## Description
I believe the partial derivative of Reynolds number with respect to mu might have an error. 

d/dmu[ (v*rho)/mu ]= -(v*rho)/(mu^2)

### Steps to reproduce issue
review:
OpenAeroStruct/openaerostruct/common/reynolds_comp.py

### Current behavior
`partials['re', 'mu'] = -inputs['rho'] * inputs['v'] / inputs['mu']**-2`

### Expected behavior
`partials['re', 'mu'] = -inputs['rho'] * inputs['v'] / inputs['mu']**2`

Please let me know if I am wrong, it happens fairly often. I don't know what effect this may have on other aspects of the code.

johnjasa:
Thanks for documenting this so well, Eric! You're indeed correct, and I've pushed a bug fix and updated the tests so they would catch this.

PR #310, opened 2020-04-21
Update OAS to OM 3.0 API

johnjasa:
## Purpose
OpenMDAO recently released version 3.0 and OpenAeroStruct was not fully compatible with its new API.
This PR updates all OAS code to be OM 3.0 compliant.
The previous PR didn't trigger a Travis build, so I'm creating this new PR.

## Type of change
What types of change is it?

- Refactoring (no functional changes, no API changes)

## Testing
Tests pass locally with OM master

## Checklist

- [X] I have run unit and regression tests which pass locally with my changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,12 +47,16 @@ install:
 
 # install openmdao and sphinx-travis
 # we only install docs dependencies on py3
+# install pyoptsparse
+- git clone https://github.com/openmdao/openmdao.git;
+- cd openmdao;
 - if [ "$PY" == "3.7" ];  then
     pip install --user travis-sphinx;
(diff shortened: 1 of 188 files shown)


johnjasa:
@nwu63 could you give this a look, please? I'm updating to OM 3.0 syntax. Also fixed #308.

nwu63:
Hey, thanks for putting in the effort to do this! Everything looks good, just a few notes:
- Would it make sense to explain either in the README or in the installation docs that Python 2 is no longer supported?
- Just to be clear, this is no longer compatible with OM 2.x right? If so, then I think this should also be made explicit either in the README or in the docs.

PR #312, opened 2020-05-02
Updated readme and installation guide

johnjasa:
## Purpose
Follow-up to our discussion about matching up OpenMDAO version and OAS version well.
This OAS version is tied to OM 3.1.0 which released today.

## Type of change
- Documentation update

## Testing
All tests pass locally with OM 3.1.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,11 +40,8 @@ install:
 - export PATH=$HOME/.local/bin:$PATH;
 
 # install openmdao and sphinx-travis
-- git clone https://github.com/openmdao/openmdao.git;
-- cd openmdao;
+- pip install `openmdao==3.1.0`
 - pip install --user travis-sphinx;
-- pip install ".[docs]";
-- cd ..;
 
 # install openaerostruct itself
 - pip install -e .;
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ install:
 - export PATH=$HOME/.local/bin:$PATH;
 
 # install openmdao and sphinx-travis
-- pip install `openmdao==3.1.0`
+- pip install openmdao
 - pip install --user travis-sphinx;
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -40,11 +40,8 @@ install:
 - export PATH=$HOME/.local/bin:$PATH;
 
 # install openmdao and sphinx-travis
-- git clone https://github.com/openmdao/openmdao.git;
-- cd openmdao;
+- pip install `openmdao==3.1.0`
```
nwu63, line 43: I think the backticks are causing Travis to error out, don't think they're needed here.

bernardopacini: Do we need to specify 'openmdao==3.1.0' or can we instead leave it without a specific version so that when openmdao is updated in the future we do not need to change our yml file to have travis catch issues?

bernardopacini, line 43: Do we need to specify 'openmdao==3.1.0' or can we instead leave it without a specific version so that when openmdao is updated in the future we do not need to change our yml file to have travis catch issues?

nwu63: I'm fine with not pinning the version and just relying on pip. That way, if anything breaks we'll know, and it's also consistent with stating that we support v3.1.0+.

bernardopacini: Request to change:
`pip install 'openmdao==3.1.0'`
To:
`pip install 'openmdao'`

johnjasa:
Thanks for the feedback! All of it makes sense. Going forward, if somebody submits a PR and it fails, know that might be because the OM API changed. Since we only Travis test on new pulls and not on OM increments, that might come up. But then we'll just change it for the latest OM.

johnjasa:
Doc build requires redbaron, let me fix that

johnjasa:
Thanks for your patience! Glad to get this out.

Issue #314, opened 2020-05-14
Possible bug in load transfer scheme


johnjasa:
## Type of issue
What types of issue is it?

- Bugfix (non-breaking change which fixes an issue)

## Description
The loads transfer scheme used in `load_transfer.py` is different than that presented in the OpenAeroStruct journal article.

In the transfer loads script, it first computes the moments at the middle of the spar elements but does not transfer the moment to the nodes using the formula of "moment transport" in that the code doesn't consider the moment arm between the node and the middle of the element.

### Steps to reproduce issue
Run any aerostructural analysis and examine the `load_transfer.py` component.

### Current behavior
The moment arms are computed from the aerodynamic centers to the structural centers.

### Expected behavior
The moment arms may need to be computed from the aerodynamic centers to the structural nodes; this theory should be double-checked as correct before implemented.

If the transfer needs correcting, the `compute()` and `compute_partials()` methods within that component would both need updating, as well as all impacted tests, namely any aerostructural test.

kanekosh:
I have fixed the bug in `load_transger.py`, though I haven't updated the tests yet so a PR is not ready. I am wondering if I can get quick comments/feedback before updating the tests. The new code can be found in [my branch.](https://github.com/kanekosh/OpenAeroStruct/commits/fix_load_transfer)
@johnjasa @shamsheersc19 

## Verification
To verify the new implementation, I did three checks.

### 1. Comparison with current implementation
I compared the aerostructural analysis of the new implementation to the current version of `mdolab/master`. I did not do any aerostructural optimization here.

Here is the comparison of aero and structural responses. The test setup is the same as `examples/run_CRM.py` except for the mesh sizes.

| Mesh size | Version | CL | CD | Tip displacement [m] | Max Von Mises [MPa]|
| --- | --- | --- | --- | --- | --- |
| `ny=5, nx=2` | current | 0.55486797 | 0.03976179 | 0.78911463 | 55.341826 |
|                        | new  | 0.57066144 | 0.03977402 | 0.30585531 | 18.679515 |
|                        | error | 2.85 %           | 0.0308 %     | -61.2 %         | -66.24 %  | 
| `ny=15, nx=5` | current | 0.44880370 | 0.03592364 | 0.32427078 | 27.272997 |
|                        | new  | 0.44994858 | 0.03593000 | 0.29448746 | 25.793820 |
|                        | error | 0.255 %            | 0.0177 %      | -9.18 %         | -5.42 %   | 
[...]

PR #318, opened 2020-06-12
Bug fixed in load transfer scheme

kanekosh:
## Purpose
This PR is to fix the issue #314. The load transfer scheme was fixed so that it is now consistent with the formulas presented in the journal paper `Jasa2018a`. Further details about the bug can be found in #314.  
Once approved, a new version 2.2.1 should be released.

I updated the followings:
- Load transfer computation and its partial derivatives
- Reference values of all aerostructural regression tests
- Documentation (not related to the bug): n2 diagram from the latest OpenMDAO 3.1.1

## Type of change

- Documentation update

## Testing
- To verify the new load transfer implementation, Sham and I performed mesh refinement studies that compare the new and old codes. We confirmed that differences in both AS analysis and optimization are small enough for reasonably fine meshes (details in #314). 
- Partial derivatives implementation was verified by the load transfer unit test, which calls OpenMDAO's `check_partials()`. 

## Checklist


nwu63:
I don't have a lot of experience with this stuff so I'll defer to Sham and John. For that one failed test, is it related to this PR? I'm wondering if we should retrain the reference value instead of just relaxing the tolerance.

kanekosh:
Yes, the failed test was about this PR. The relative error was slightly above `1.e-5` for one of the optimization tests. Though I haven't confirmed, I think it was because of the version difference of a `scipy` optimizer between my local and Travis build.

nwu63:
Oh sorry I got confused, I thought you only changed the tolerance, but in fact you had already re-trained the values, but it somehow failed on Travis, due to some discrepancy between Travis and your local machine. Understood, I'm fine with this PR then.

shamsheersc19:
> @johnjasa and @shamsheersc19 do you feel that this deserves a minor version increment instead of a bugfix release?

@nwu63 I think this deserves a minor version increment.

nwu63:
> @nwu63 I think this deserves a minor version increment.

If that is the case, then please increment the version and make a new release. I think @kanekosh should do this since he did the PR. Feel free to message me if you have any questions on how you should do this.

shamsheersc19:
> If that is the case, then please increment the version and make a new release. I think @kanekosh should do this since he did the PR. Feel free to message me if you have any questions on how you should do this.

I misunderstood the versioning terminology. Proceeding with a patch release. Thanks!

PR #319, opened 2020-06-15
Fixed docs (n2 diagram)

kanekosh:
## Purpose
Fixed the documentation to accommodate a new format of n2 diagram in `Aerodynamic Optimization Walkthrough`. 

## Type of change
- Documentation update

## Testing
Local build succeeded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,8 +12,9 @@
 *.0
 *.o
 *.mod
-docs/_build/
-docs/*.rst
+openaerostruct/docs/_build/
+openaerostruct/docs/*.rst
+openaerostruct/docs/_srcdocs/
 *.a
 *.msg
 src/adjoint/tempReverse/
--- a/openaerostruct/docs/_n2html/generate_n2_n2.html
+++ b/openaerostruct/docs/_n2html/generate_n2_n2.html
@@ -1,10 +1,3 @@
-<!doctype html>
-<html >
-<head >
-<title>OpenMDAO Model Hierarchy and N2 diagram</title>
-
-<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
-
 <style type="text/css">
 input:focus,
 select:focus,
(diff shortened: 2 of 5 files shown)


kanekosh:
Yes this was the only one.

nwu63:
I think it would be good to get @johnjasa to double check this, as I'm not really familiar with how sphinx works with Travis/GitHub Pages for building/deploying docs.

johnjasa:
Thanks for this fix, Shugo. I like how you're using the `embed-n2` directive from OM now.

I tried building the docs locally in the `openaerostruct/docs` folder by using `make html`, but they failed with this message:

```Exception occurred:
  File "/mnt/c/Users/jjasa/git/OpenMDAO/openmdao/docs/_exts/embed_n2.py", line 56, in run
    raise IOError('File does not exist({0})'.format(np))
OSError: File does not exist(/mnt/c/Users/jjasa/git/OpenAeroStruct/openaerostruct/docs/openaerostruct/docs/aero_walkthrough/generate_n2.py)
```

Basically, the path was wrong to the `generate_n2.py` file with duplicate `openaerostruct/docs` parts there. Does the local build work for you? This might be something from before and not from your changes. I'll keep looking at it a bit.

johnjasa:
Don't try tackling it - I think I've got it. Can fix it locally in just a bit and I'll push to a branch.

kanekosh:
Thanks John for checking this! The local build actually fails for me as well. When we build locally in `openaerostruct/docs`, the path origin is `openaerostruct/docs`. However with Travis, it seems the origin is `OpenAeroStruct` and Travis test fails in that case.
But I guess I'm missing something as your `embed-code` works but `embed-n2` does not.

johnjasa:
I see, that's right on the different origin directories. Do you think the paths I changed in `aero_walkthrough.rst` will work for Travis, then? They work locally, but it seems there might be different behavior.

Does us [including the docs path here](https://github.com/mdolab/OpenAeroStruct/blob/master/openaerostruct/docs/conf.py#L17) help that at all? Maybe I'm misinterpreting what that does, it's been a while.

johnjasa:
I guess what it comes down to is that it'd be nice if the docs can be made both on Travis and locally, in my opinion. Do you think that's worth sorting out, or should we say Travis-only is okay?

kanekosh:
I think it will fail, at least for `embed-code`. Actually my initial commit was similar and it failed.
But since the paths seem to behave differently between `embed-code` and `embed-n2` when built locally, I don't know what will happen with Travis.
I agree, we should be able to check if Travis works by doing `make html` locally, otherwise the same thing like this will happen and bother reviewers again.

johnjasa:
Oof, I should've scrolled up and checked that. Thanks for clarifying. I've pushed a fix to the `.travis.yml` to change directories, hopefully it works.

johnjasa:
Neat, the build looked okay on Travis. All good in my book!

johnjasa:
Wait, Coveralls looks wacky still: https://coveralls.io/jobs/63931317
I tried to fix it but I'm not certain if that'll do it. I see it was reported as 0% earlier, too.
@kanekosh, do you have any ideas for that?

kanekosh:
`make html` works on my local too, thanks!
I think Coveralls will work fine this time. When I first passed Travis build (but not locally) in `
94c9498`, where Travis was working on `OpenAeroStruct` dir, there was no problem.

johnjasa:
Okay, great! If Coveralls is still weird, I'll leave it up to you to handle. :)

nwu63:
Is this PR ready to be merged? Happy to do so as long as you guys ironed out the issues @johnjasa @kanekosh

kanekosh:
It's ready!

PR #332, opened 2020-08-27
Edited requirement packages; Revived FFD tests

### LINKED ISSUE #306: Update setup.py
## Type of issue
What types of issue is it?
_Select the appropriate type(s) that describe this issue_

- New feature (non-breaking change which adds functionality)

## Description
Update setup.py to automatically handle Python package dependencies. Should also pin specific versions.

kanekosh:
## Purpose
- Added packages required by OAS to `setup.py`: `openmdao 3.2+, numpy, scipy, matplotlib, six`. Accordingly, commands to install these were removed from `.travis.yml`. Note that packages that are installed in the travis file but don't appear in `setup.py` are the ones only required for testing/building documents, which users don't need on their local machine.  (closes #306)
- In addition, I made a quick syntax fix/formating of examples. 

## Type of change
- Documentation update

## Checklist


nwu63 (dismissed):
Looks good. Would be interesting to see if parallel `testflo` execution will reduce the Travis run time.

kanekosh:
Thanks! But Travis build hasn't passed yet, I think my last commit fixes it but need to check.
@nwu63 Is it possible to let Travis run while the PR is still a Draft? (I don't want to spam reviewers, which I already did this time...)

nwu63:
I'm not sure actually, I think it's possible to just test this on your own Travis though, with branch builds from your name space.

As for the error, it's because `setup.py` is relying on `numpy.distutils`. I'm actually not sure why that's the case, maybe a heritage from when we had Fortran code in the package and needed f2py to run the build similar to `pyoptsparse`. If I'm not mistaken that can be removed now. @johnjasa does any of this ring a bell?

kanekosh:
- I replaced `numpy.distutils.setup` with `setuptools.setup` as recommended [here](https://docs.python.org/3/library/distutils.html). I'm not quite sure what the difference is.  
- Testflo ran in parallel with petsc4py installed, but it didn't speed up, probably due to an Open MPI issue ([build log is here](https://travis-ci.com/github/mdolab/OpenAeroStruct/builds/181578695)). Even if the test speeded up, petsc installation time cancels out (actually makes it worse), so I ended up not doing parallel testing.

nwu63:
The difference is that `numpy.distutils` is able to facilitate building a project that requires compiling Fortran code, then wrapping with `f2py`. This is what we use in pyOptSparse to build against SNOPT for example, without an intermediate build step involving makefiles. OpenAeroStruct used to have Fortran codes and I assume this is why it was there, but since all that were removed in v2 we no longer need it.

That's interesting, do you know why petsc4py was needed? There is no PETSc in the Travis environment so the petsc4py package wouldn't have done anything. In fact mpi4py is not needed either but I'm not 100% sure on that.

kanekosh:
You're right, it didn't need `petsc4py`. It was just my misunderstanding coming from the OpenMDAO's warning on missing petsc4py, which I think doesn't matter for testflo's parallel capability. Same for `mpi4py`, expect `travis-sphinx` requires it for some reason (which I'm not sure).
It looks testflo is running in parallel, but the wall time is not improved from the serial, so something is going wrong... 
Anyway, I think it's ready to be merged this time, sorry for a lot of notifications :(

.travis.yml
```diff
@@ -1,10 +1,12 @@
...
 - linux
 
-language: generic
+language: python
 
+python:
+  - 3.7
 env:
-- PY=3.7
+  - PARALLELIZE=true
```
nwu63, line 9: Why is this needed?

.travis.yml
```diff
@@ -24,9 +26,12 @@ before_install:
 - export PATH=/home/travis/miniconda/bin:$PATH;
 
 install:
-- conda install --yes python=$PY numpy scipy nose sphinx mock swig pip matplotlib numpydoc;
+- conda install --yes python=$PY nose sphinx mock swig pip numpydoc;
 - pip install --upgrade pip;
-- pip install mpi4py;
+- pip install mpi4py
```
nwu63, line 31: I'm not convinced why we need `mpi4py` since AFAIK we are not using any parallel capability of OpenMDAO. Not sure why there was an error during docs build, that looks like an internal OpenMDAO error to me.

kanekosh:
Yeah this is a good opportunity to fix them. Let me check which tests are being skipped, and if they are necessary.

kanekosh:
Out of 7 skipped tests, 4 are due to the lack of analytical derivatives, so they cannot be added right away. The other 3 are FFD tests, which were skipped because they require pygeo (therefore mpi4py and pyspline). I think these can be added to Travis build tests now.  
I don't want to add pygeo/pyspline to the installation requirements since it needs mpi4py and some fortran complications; just want to keep OAS a pure python package for most of the users.

nwu63:
Oh I see, that makes sense. Are FFDs officially supported by OpenAeroStruct? I know John implemented them before but are they documented/tested at all? Yes that makes  sense, optional features should not have mandatory dependencies. It is possible to specify different sets of dependencies in `setup.py`, this is what OpenMDAO does since their docs/tests have different requirements. Maybe that's something we can do also.

Actually, I just realized, can you try removing all the testing dependencies in setup.py, and instead do `pip install openmdao[docs]`? If you check their `setup.py` file [here](https://github.com/OpenMDAO/OpenMDAO/blob/master/setup.py) you can see that all the testing dependencies are listed under `docs` which will be installed.

Still not sure about this `mpi4py` issue though, I'm trying to reproduce it locally. But travis-sphinx is something specific to Travis so not sure if that will work.

kanekosh:
There are two regression tests (aero and aerostructural optimization) that involve FFD, also a page in the documentation, so I think it's fine.

`openmdao[docs]` worked great!

Still not sure how to fix the mpi4py issue in travis-sphinx, FFD tests (pygeo and pyspline) requires mpi4py installation either way, so they hide the issue (which is not great)

nwu63:
Right, I still think everything we test here should not need `mpi4py` since we don't do anything in parallel, right? We don't use any MPI capability in OpenMDAO and we also don't use parallel capabilities in pygeo/pyspline. So maybe we should actually remove mpi4py as requirements for those packages since they _can_ be run on one proc, such as when used with OpenAeroStruct.

Not sure how to fix the travis-sphinx issue though, still looks like a mock MPI comm issue within OpenMDAO but I'm not sure. Maybe we can talk to the developers?

kanekosh:
Correct, OAS tests don't require mpi4py. I'll submit an issue, the rest should be pygeo maintainers' call.

We should do so, I'll organize the info and post in the OM public channel later.

kanekosh:
Thanks @nwu63 too! Will contact the OM team. I don't think pygeo can remove mpi4py since they need it for MACH. For my OAS case, ModuleNotFoundError occurred when I import pygeo which imports mpi4py internally. 

I submitted an issue to requests to add `mpi4py` as a requirement in their `setup.py`, because mpi4py is necessary even when running pygeo in serial. (Or they could add try/except for mpi4py import, but it's up to them.)

PR #337, opened 2020-08-31
Fixed multipoint tests

### LINKED ISSUE #284: Fix multipoint tests to be correct
Some of the example scripts for multipoint are wrong in that there are unnecessary geometry groups added.
Here's an example that needs to be fixed:
https://github.com/mdolab/OpenAeroStruct/blob/master/openaerostruct/tests/test_multipoint_aero.py
There may be others that need fixing as well, please check all multipoint files.

This one is correctly implemented:
https://github.com/mdolab/OpenAeroStruct/blob/master/openaerostruct/tests/test_multipoint_wingbox_aerostruct.py

It's all about where we add the geometry group for each lifting surface.

kanekosh:
## Purpose
Some multipoint tests (which also appear in the docs) were not actually multipoint, but they were multi-aircraft/ morphing. This is because the geometry groups were added under each aero points repeatedly and individually, although the geometry should be consistent between all aero points.

This PR fixes #284, by removing the geometry groups from each aero point, adding them outside the aero point groups, and connecting the single geometry to each aero point.  
I also needed to update the reference values in `test_multipoint_aero.py`, because the optimization problem it solves is changed.

## Type of change

## Checklist
- [ x ] I have run unit and regression tests which pass locally with my changes

### DIFF
--- a/openaerostruct/tests/test_multi_aero_ffd.py
+++ b/openaerostruct/tests/test_multi_aero_ffd.py
@@ -1,72 +1,68 @@
-from openmdao.utils.assert_utils import assert_rel_error
 import numpy as np
 import unittest
 
 try:
     import pygeo
+
     pygeo_flag = True
-except:
+except ModuleNotFoundError:
     pygeo_flag = False
+
 
 @unittest.skipUnless(pygeo_flag, "pyGeo is required.")
 class Test(unittest.TestCase):
-
     def test(self):
 
         from openaerostruct.geometry.utils import generate_mesh, write_FFD_file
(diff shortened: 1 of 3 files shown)


kanekosh:
@galenng525 since you were working on multipoint cases on OAS, it'd be great if you can take a look and check if the code I fixed makes sense to you.

kanekosh:
Yes I renamed because `multi` is ambiguous (could be either multipoint or multi-surface). I used `git mv` so it should keep tracking diffs, but I guess something weird is happening when squashing.  
Would you refer to my second commit `fixed multipoint tests` to see the diffs?

PR #339, opened 2020-08-31
Updated README

kanekosh:
## Purpose
- Updated the link of the OAS journal paper in the README. Also added the link for Sham's wingbox paper.
- Cleaned up the root directory: moved `OAS_xdsm.pdf` to the `docs`, and `example_deriv_check.py` to the `example` folder, because I don't see the reason to keep them at the root.
- Added the XDSM figure to the main page of the docs. 

## Type of change

- Documentation update
- Maintenance


shamsheersc19:
Just replaced the wingbox paper link to ResearchGate (for the stats).

PR #327, opened 2020-08-01
Adding an example for making custom meshes

shamsheersc19:
## Purpose
Adding examples for how to make a custom mesh based on popular demand.
For #246 and #283.

## Type of change
What types of change is it?
- Documentation update

## Testing
Try the new example and see if it helps you make your own mesh.

## Checklist
_Put an `x` in the boxes that apply._

- [x] I have run unit and regression tests which pass locally with my changes
- [ ] I have added new tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation

PR #343, opened 2020-11-09
Promote beta (sideslip angle) to the top of AerostructGroup

### LINKED ISSUE #342: The input "Beta" has not been promoted through the AerostructPoint group.
## Description
The input "Beta" has not been promoted through the AerostructPoint group. It should be accessible at the top just like alpha without requiring connections to multiple places inside the group.

### Steps to reproduce issue
_Please provide a minimum working example (MWE) if possible_

1. Run test_aerostruct.py in the debugger and stop after setup.
2. prob['AS_point_0.beta']

### Current behavior
prob['AS_point_0.beta']
...
builtins.KeyError: '<model> <class Group>: Variable "AS_point_0.beta" not found.'

### Expected behavior
prob['AS_point_0.beta']
array([0.])

### Code versions
_List versions only if relevant_
- Python
- …

Kenneth-T-Moore:
## Purpose
I was running a case where I needed to compute the derivative of an OpenAerostruct model with respect to beta, and found that it wasn't promoted. It was actually partially promoted in a couple of spots, but some important ones were missing.

## Type of change
What types of change is it?

closes #342

- Breaking change (non-backwards-compatible fix or feature)
Not a major "breaking" change, but since version 3.2 of OpenMDAO, auto-indepvarcomps are automatically created for unconnected inputs. To prevent ambiguity, if inputs with different units are promoted together, OpenMDAO raises an error. You can work around this by adding that input to your top level indepvarcomp, but a more correct way is to use `group.set_input_defaults` to define the source units (but use of this will tie you to OpenMDAO 3.2 or greater.)

See test_aerostruct_analysis_compressible.py.  This will only occur when using the compressible aerodynamic analysis, as the transformation to the wind-frame declares the input beta in radians instead of degrees.

Note also that if the user omits some of the other variables from the indpevarcomp at the top of the model (for example, alpha), they also get this same error. You may want to consider making the mods to OAS to work with the latest OpenMDAO, but that is a separate issue.

## Testing
Quick test added to an existing test.

## Checklist

- [NA] I have added necessary documentation

### DIFF
--- a/openaerostruct/integration/aerostruct_groups.py
+++ b/openaerostruct/integration/aerostruct_groups.py
@@ -127,7 +127,7 @@ class CoupledPerformance(om.Group):
 
         self.add_subsystem('aero_funcs',
             VLMFunctionals(surface=surface),
-            promotes_inputs=['v', 'alpha', 'Mach_number', 're', 'rho', 'widths', 'cos_sweep', 'lengths', 'S_ref', 'sec_forces', 't_over_c'], promotes_outputs=['CDv', 'L', 'D', 'CL1', 'CDi', 'CD', 'CL'])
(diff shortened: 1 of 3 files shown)


kanekosh (approved):
Thank you!
We already require OpenMDAO 3.2+ (and did a `group.set_input_defaults` fix in #327), so this should be good.

PR #351, opened 2021-01-30
Migrate to Github Actions

kanekosh:
## Purpose
Migrate CI/CD tool from Travis to Github Actions.
- `oas_pr.yml` is triggered when a PR is created. This runs tests and builds docs (but does not deploy it).
- `oas_push.yml` is triggered when a PR is merged to master (i.e., pushed to master). In addition to the tests and doc builds, it deploys the docs to gh-pages.

## Type of change
What types of change is it?
- Other (maintenance update)

## Checklist


kanekosh:
Just for future reference:
This extension: https://github.com/marketplace/actions/sphinx-build looked helpful, but for some reason it required re-installing everything (openaerostruct, openmdao, mpi4py etc...) in its own block using `pre-build-command`.
Since these packages are already installed in earlier blocks (for tests), I decided to not use it and directly build the docs via `make html`.

kanekosh:
I believe it's working now. If it looks good, I will remove the travis file before merging.
It looks to me that GH Actions is properly blocking the merge (as travis was doing so), but I didn't any settings for that, so I am a bit worried.

.github/workflows/oas_pr.yml
```diff
@@ -0,0 +1,58 @@
...
+    # build docs (but not deployed)
+    - name: Build docs
+      run: |
+        cd openaerostruct/docs
+        make clean
+        make html
+
+    - name: Run coveralls
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
nwu63, line 55: Where is this stored? I don't see any secrets defined in the repo settings. I guess the token is needed to be able to comment on the PR after? Does it need any token to submit to Coveralls?

nwu63 (commented):
Looks fine. My main question is whether it's possible to combine these into a single file. I don't like the code duplication in two files that are largely the same.

.github/workflows/oas_pr.yml
kanekosh: I didn't understand very well, but `GITHUB_TOKEN` is a secret automatically created, so we don't need to define this by ourselves.
https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

And we need this token to submit to coveralls.

kanekosh:
> Looks fine. My main question is whether it's possible to combine these into a single file. I don't like the code duplication in two files that are largely the same.

Agreed, will take a look, I believe it should be possible.

.github/workflows/oas_pr.yml
nwu63: I see, that makes sense thanks.

nwu63:
One last thing: I believe we want `coveralls` to run on the PR build also, so that we can monitor the testing coverage and make sure it doesn't decrease by much. On Travis, the `after_success` stage is run regardless of whether it's a push to master or PR build, and I guess `travis-sphinx deploy` just failed on PR build since it requires a secret token (set on Travis website) to deploy.

kanekosh:
I believe it's good to go! I unified two files and added if statements so that docs deployment and coveralls are only called when we push to the master branch. I tested the if statements [here](https://github.com/mdolab/OpenAeroStruct/actions/runs/526413579) and it worked.

I also removed the Travis file.

nwu63:
Awesome stuff! Sorry to be picky but I really wanna make sure we're careful here. Did you ever test that this is working as expected? i.e. push a commit that deliberately broke something and check that GHA fails.

kanekosh:
I deliberately broke one of the tests, and now GHA fails as it is supposed to.
The documentation build is a bit tricky, though. GHA fails when `make html` fails, for example when we didn't specify the correct dependencies on `requirements.txt`. However, when `make html` finishes with some warnings, GHA does not detect and block that. This happens when embedded python scripts have some errors. In such a case, the docs do have embedded code on them, but since python execution failed, we don't have the python output (like optimization result printout) on the docs.

I am not sure if we can deal with this at the GHA level.

nwu63:
Yeah that seems like something to figure out with the OpenMDAO toolchain for building docs. We can investigate that in the future. Happy to merge now.

nwu63 (approved):

PR #357, opened 2021-02-19
Update setup.py for OM reqs

shamsheersc19:
## Purpose
Update setup.py to restrict OpenMDAO range.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setup(name='openaerostruct',
     },
     # TODO: add versions?
     install_requires=[
-        'openmdao[docs]>=3.2',
+        'openmdao[docs]>=3.2, <=3.6.0',
         'numpy',
         'scipy',
         'matplotlib',


nwu63:
Haha why did you open 3 separate PRs? Anyway let's merge this first so the other ones can pass.

nwu63 (approved):

shamsheersc19:
@nwu63 lol I was trying the directly editing in GitHub approach. I might be missing something, but it doesn't seem like you can edit multiple files and add them to the same branch?

nwu63:
> @nwu63 lol I was trying the directly editing in GitHub approach. I might be missing something, but it doesn't seem like you can edit multiple files and add them to the same branch?

No you can, there are two ways on the web (without using git locally)
1. Do as you did now, and create a PR with one edit. Then go back into the same branch, edit and push directly to the same branch.
2. Create a new branch from master, then make multiple commits directly, and make a PR at the end.

PR #359, opened 2021-03-12
Rcharayr custom mesh

RemyCharayron:
Handle custom mesh with dihedral distribution.
Run files for IAI Scout and AAI Shadow 200 to compare structural mass.


RemyCharayron:
DO NOT TAKE THIS INTO ACCOUNT THAT WAS AN ERROR

PR #361, opened 2021-07-30
fixes for OM version update

### LINKED ISSUE #354: Errors with OpenMDAO 3.7.0
## Description
Errors with OpenMDAO 3.7.0:
\openmdao\core\component.py", line 1347, in _declare_partials
    raise ValueError('{}: No matches were found for wrt="{}"'.format(self.msginfo,
ValueError: 'AS_point_0.wing_perf.aero_funcs.liftdrag' <class LiftDrag>: No matches were found for wrt="L"

I don't think it's worth updating OAS every time OM changes, which is quite frequent. I think we should just be specific about which version people should use (i.e., OM 3.2 to 3.6) 

### Steps to reproduce issue
Use OpenMDAO 3.7.0 with any example script in the examples dir.

kanekosh:
## Purpose

Minor fixes for the recent OpenMDAO versions (now works with OM up to 3.10.0). Most of the fixes are the removal/edits of the partial derivative tests on the components we don't have analytic derivatives. Closes #354 .

To build the sphinx documentation, we need OpenMDAO<=3.9.2 because OpenMDAO migrated their docs to jupyter notebooks at 3.10. I made a few changes to `setup.py` and Github action workflow for this.

## Type of change
What types of change is it?

- Maintenance

## Checklist


nwu63:
Is this PR ready for review?

kanekosh:
Not yet, I’ll request review when it’s ready

kanekosh:
Ready for review now!

openaerostruct/tests/test_aero_morphing_multipoint.py
```diff
@@ -1,4 +1,4 @@
-from openmdao.utils.assert_utils import assert_rel_error
+from openmdao.utils.assert_utils import assert_near_equal
```
nwu63, line 1: I personally prefer the asserts from `numpy`, i.e. `np.testing.assert_all_close`, since it's good for both large and small numbers. But if the math behind this is okay then I'm fine with that also.

openaerostruct/tests/test_aerostruct.py
```diff
@@ -123,10 +123,10 @@ def test(self):
         prob.driver = om.ScipyOptimizeDriver()
         prob.driver.options['tol'] = 1e-9
 
-        recorder = om.SqliteRecorder("aerostruct.db")
-        prob.driver.add_recorder(recorder)
-        prob.driver.recording_options['record_derivatives'] = True
-        prob.driver.recording_options['includes'] = ['*']
+        # recorder = om.SqliteRecorder("aerostruct.db")
```
nwu63, line 126: If not needed we should remove the code

setup.py
```diff
@@ -6,6 +6,19 @@
     r"""__version__ = ["']+([0-9\.]*)["']+""",
     open('openaerostruct/__init__.py').read(),
 )[0]
+
+optional_dependencies = {
+    'docs': ['openmdao[docs]==3.9.2'],
```
nwu63, line 11: Do we need to pin this version here?

setup.py
```diff
@@ -6,6 +6,19 @@
     r"""__version__ = ["']+([0-9\.]*)["']+""",
     open('openaerostruct/__init__.py').read(),
 )[0]
+
+optional_dependencies = {
+    'docs': ['openmdao[docs]==3.9.2'],
+    'test': ['testflo>=1.3.6']
+}
+
+# Add an optional dependency that concatenates all others
```
nwu63, line 15: Is this mentioned/used anywhere? Might be good to update docs or the GitHub Actions job to use this if suitable. If not, then don't worry about it, I just don't like adding code that's not actually used.

nwu63 (changes requested):
Just some minor comments. I think we will want to update the README to update the OpenMDAO version listed there.

setup.py
kanekosh: Now I specify the docs dependencies (like numpydoc) directly instead of relying on openmdao

setup.py
kanekosh: We use `pip install .[all]` in Github Action. Also I'll add a note to docs or README

setup.py
kanekosh: OpenMDAO has to be 3.9.2 or older because they migrated their docs to jupyter notebooks at 3.10.0. We cannot build the OAS docs with OpenMDAO 3.10.0.

nwu63: Ah okay, makes sense

setup.py
nwu63: Cool sounds good.

openaerostruct/tests/test_aero_morphing_multipoint.py
kanekosh: `assert_near_equal` checks relative errors so it should be good I think.

nwu63 (dismissed):

kanekosh:
Updated README and docs/installation.rst.

I also made minor docs changes so that we don't need to run all optimizations during the build process which takes time. Now the docs only show the scripts (no outputs) for advanced examples. It still shows outputs for the fundamental examples (aero/structure/aerostructural walkthrough) .

nwu63 (approved):

nwu63:
Do we want to make a new release?

openaerostruct/aerodynamics/eval_mtx.py
```diff
@@ -266,8 +266,7 @@ def setup(self):
 
             # It's worth the cs cost here because alpha is just a scalar
             self.declare_partials(vel_mtx_name, 'alpha', method='cs')
-
-            self.set_check_partial_options(wrt='*', method='cs')
+            self.set_check_partial_options(wrt='alpha', method='fd')
```
bernardopacini, line 269: What is the motivation here to switch to fd?

bernardopacini (approved):
This all looks good to me, my only questions are about using finite-difference for partial derivative checks. Is this a change from before?

openaerostruct/aerodynamics/eval_mtx.py
kanekosh: It computes the partials using CS and checks them using CS which doesn’t make sense. The latest version of OpenMDAO raise an error in such case. I changed to check with FD to avoid the error (essentially it does nothing though)

kanekosh:
@nwu63 I’ll do black/flake8 changes next week (totally forgot about it) and release a new version then

nwu63:
Merging since you also approved it @bernardopacini, hopefully the question you had was addressed. Going to merge now.

PR #362, opened 2021-08-18
Formatting (flake8 and black)

kanekosh:
## Purp
Formatting by flake8 and black.

TODO
[] update github action yml
[] update PR template
[] bump version

## Type of change
What types of change is it?

- New feature (non-breaking change which adds functionality)
- Breaking change (non-backwards-compatible fix or feature)

## Checklist


openaerostruct/examples/run_aerostruct_uCRM_multipoint.py
```diff
@@ -533,9 +536,6 @@
 # =======================================================================================
 # =======================================================================================
 
-from openmdao.utils.assert_utils import assert_rel_error
```
nwu63, line 316: How does this part of the code work? The example runscript is also used for unit testing? Is this being tested now?

openaerostruct/geometry/ffd_component.py
```diff
@@ -6,7 +6,7 @@
 
 try:
     from pygeo import DVGeometry
-except:
+except ImportError():
```
nwu63, line 9: I think you want `ImportError` without the `()`

openaerostruct/tests/test_aerostruct_ffd.py
```diff
@@ -3,11 +3,12 @@
...
 from openaerostruct.utils.constants import grav_constant
 
+# check if pygeo is available
 try:
-    import pygeo
+    import pygeo  # noqa: F401
 
     pygeo_flag = True
-except:
+except ImportError():
```
nwu63, line 11: Same here

openaerostruct/utils/plot_wing.py
```diff
@@ -12,36 +12,28 @@
...
-from matplotlib import cm
-import matplotlib.animation as manimation
 
-from traceback import print_exc
+major_python_version = sys.version_info[0]
 
-#####################
-# User-set parameters
-#####################
+if major_python_version == 2:
```
nwu63, line 31: I don't think we support py2 anymore right?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,24 +1,26 @@
 ## Purpose
-Explain the goal of this PR, if it addresses an existing issue be sure to link to it. Describe the big picture of your changes here, perhaps using a bullet list if multiple changes are done to accomplish a single goal.  If it accomplishes multiple goals, it would be best to create separate PR's for each. 
+Explain the goal of this PR, if it addresses an existing issue be sure to link to it. Describe the big picture of your changes here, perhaps using a bullet list if multiple changes are done to accomplish a single goal. If it accomplishes multiple goals, it may be best to create separate PR's for each.
```
nwu63, line 2: If the default MDO Lab templates [here](https://github.com/mdolab/.github/tree/master/.github) are okay, then we can just delete this file and GitHub will automatically use the default organizational template. Same with the ISSUE template.

nwu63 (changes requested):
See below for minor comments. There are two things I'd like to change for the flake8/black setup:
1. Run them as separate tests, rather than sequentially under `build`. That way, we get parallelization, and the status badge will show whether the tests or black/flake8 failed. So having 3 separate jobs make sense to me.
2. I would prefer duplicating as little of the flake8 config as possible. This can be done by deferring as much configuration as possible to the organization config file [here](https://github.com/mdolab/.github/blob/master/.flake8), then putting the rest in the repo-specific file. `flake8` has some handy CLI options to pass in two config files, I can help you with that after 1) is done.

openaerostruct/utils/plot_wing.py
kanekosh: That's right

.github/PULL_REQUEST_TEMPLATE.md
kanekosh: Yes, will delete the template from OAS repo then.

openaerostruct/examples/run_aerostruct_uCRM_multipoint.py
kanekosh: For some reason this example file includes a regression test in the latter part of the script, which was not actually tested during the testflo call. I removed this test because what it overwraps with other regression tests we already have.

nwu63: Yeah that sounds good, this seemed confusing from a user perspective.

nwu63 (commented):
Sorry I don't think I was clear about the CI jobs. I would prefer one file, but multiple _jobs_, i.e.
```
jobs:
  build_and_test:
    ....
  black:
    ....
  flake8:
    ....
```
Each of these jobs will have multiple _steps_ as before, but keeping them in one file makes the GitHub Actions outcomes easier to read on the Actions tab. See for example [this GHA script](https://github.com/mdolab/niceplots/blob/master/.github/workflows/niceplots.yml).


Also, there is no reason to install OAS/dependencies for flake8/black, since they're just linting the source code and not running it.

kanekosh:
That makes more sense, I think all fixes are done now.

nwu63 (dismissed):
Just some very minor details, overall this looks great.

nwu63 (approved):

nwu63:
Did you want to wait for @bernardopacini's feedback or do you want me to merge the PR? It looks like the repo is set up to only require 1 PR review

kanekosh:
I think we can merge now (I can also do that so I will)

PR #363, opened 2021-08-23
add OM-OMPI env var

kanekosh:
## Purpose
Fixes the documentation build.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/workflows/oas.yml
+++ b/.github/workflows/oas.yml
@@ -106,6 +106,7 @@ jobs:
     # build docs
     - name: Build docs
       run: |
+        export OPENMDAO_USE_MPI=False
         python -m pip install --upgrade pip wheel
         pip install .[docs]
         cd openaerostruct/docs


nwu63 (approved):
Perhaps it would be a bit more explicit if we used the `env` syntax as done [here](https://github.com/kanekosh/OpenAeroStruct/blob/fix-docs/.github/workflows/oas.yml#L47-L48)? I have no strong feelings either way.

kanekosh:
The variable `OPENMDAO_USE_MPI=False` worked locally but does not work on GH Action

kanekosh:
Fixed now, ready to merge

.github/workflows/oas.yml
```diff
@@ -105,6 +105,8 @@ jobs:
 
     # build docs
     - name: Build docs
+      env:
+        OPENMDAO_USE_MPI: False
```
nwu63, line 109: Do we know if this is still needed?

PR #374, opened 2022-01-30
Please pass the post-merge build

eytanadler:
## Purpose

Adjusts partial derivative checking tolerances to fix the build that was merged previously. Also addresses a couple other sources of warnings including

1. Changing `value` option to `val` in OpenMDAO `ExecComp`
2. Adding units to `IndepVarComp` outputs in tests
3. Updating `assert_rel_error` calls to `assert_near_equal`
4. "Flatting" variables (convert scalar array to a floating point variable) before creating a NumPy array instead of after the array is created

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

- [ ] I have run `flake8` and `black` to make sure the code adheres to PEP-8 and is consistently formatted


openaerostruct/aerodynamics/tests/test_geometry.py
```diff
@@ -2,7 +2,7 @@
 import numpy as np
 
 import openmdao.api as om
-from openmdao.utils.assert_utils import assert_check_partials, assert_rel_error
+from openmdao.utils.assert_utils import assert_check_partials, assert_near_equal
```
nwu63, line 5: Why the switch?

openaerostruct/aerodynamics/pg_scale.py
```diff
@@ -205,17 +205,17 @@ def compute_partials(self, inputs, partials):
         rotational = self.options["rotational"]
 
         M = inputs["Mach_number"]
-        betaPG = np.sqrt(1 - M ** 2)
-        fact = np.array([1.0, betaPG, betaPG], M.dtype).flatten()
-        fact_norm = np.array([betaPG, 1.0, 1.0], M.dtype).flatten()
+        betaPG = np.sqrt(1 - M ** 2)[0]
```
nwu63, line 208: All OM inputs are cast to ndarray right? I don't like the `[0]`, it's probably better to use `np.asscalar()`

openaerostruct/aerodynamics/pg_wind_rotation.py
```diff
@@ -166,10 +166,10 @@ def compute(self, inputs, outputs):
         alpha = inputs["alpha"]
         beta = inputs["beta"]
 
-        cosa = np.cos(alpha)
-        sina = np.sin(alpha)
-        cosb = np.cos(beta)
-        sinb = np.sin(beta)
+        cosa = np.cos(alpha)[0]
```
nwu63, line 169: A better implementation here would be not to fetch these as scalars, and manually write the rotation matrix. Instead, use an existing implementation such as [scipy.spatial.transform](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html). Could be done later though, I just feel this is kind of silly

openaerostruct/aerodynamics/tests/test_geometry.py
```diff
@@ -97,7 +97,7 @@ def test_derivs_wetted(self):
         prob.run_model()
 
         check = prob.check_partials(compact_print=True, step=1e-7)
-        assert_check_partials(check, atol=5e-5, rtol=1e-5)
+        assert_check_partials(check, atol=7.5e-5, rtol=1e-5)
```
nwu63, line 100: Do we need a better step to get more accurate FD values? If we don't need to relax the tolerance we should try not to

openaerostruct/aerodynamics/tests/test_geometry.py
eytanadler: `assert_rel_error` was deprecated in OpenMDAO 3.1.0 (now on 3.16.0)

openaerostruct/aerodynamics/pg_scale.py
eytanadler: I didn't know `asscalar` existed, I agree that's a better looking solution

openaerostruct/aerodynamics/pg_wind_rotation.py
eytanadler: That makes sense, I don't want to mess with their implementation too much because I haven't worked with these components much. I'll make the switch to `np.asscalar()` (these will always be of size 1 because the input shape is defined as that size).

openaerostruct/aerodynamics/tests/test_geometry.py
eytanadler: This was something I was seeing locally on my computer, but since it's not a problem in the build I'll undo it. For some reason when I run `python -m pytest` on the full repo, this check fails. But when I run only this test (`python -m pytest openaerostruct/aerodynamics/tests/test_geometry.py`), it passes. I can't explain it

openaerostruct/aerodynamics/tests/test_geometry.py
nwu63: Ah okay. Their docs should probably state that then, it wasn't obvious to me. What's the difference between the two?

openaerostruct/aerodynamics/pg_wind_rotation.py
nwu63: Yeah this is fine, just something to keep in mind for the future.

openaerostruct/aerodynamics/tests/test_geometry.py
nwu63: Yeah no idea. Probably different numpy versions or something.

openaerostruct/aerodynamics/tests/test_geometry.py
eytanadler: It looks like it has something to do with OpenMDAO's internal test case structure. If you wanna look at the source code, [see here](https://github.com/OpenMDAO/OpenMDAO/blob/cc3b5c3b7dca870005cfc1be761746f09a26ff67/openmdao/utils/assert_utils.py). The deprecation is listed [here](https://github.com/OpenMDAO/OpenMDAO/blob/302cd32109d4450de7f9f0a1de04278516cecfcb/release_notes.md) in version 3.1.0. Other than the deprecation warning, this is the only place I've found it.

nwu63: I just diffed the code and they are very similar, so I'm happy with the change.

nwu63 (approved):

PR #375, opened 2022-01-31
Fixed random seeds

kanekosh:
## Purpose
Fixed all random seeds so that all tests will be reproducible. Also, replaced `np.random.seed()`with `np.random.default_rng()`, which is a new best practice.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,14 +69,14 @@ Shamsheer S. Chauhan and Joaquim R. R. A. Martins, “Low-Fidelity Aerostructura
 
 Version Information
 -------------------
-This version of OpenAeroStruct requires [OpenMDAO](https://github.com/OpenMDAO/openmdao) (versions 3.2.0 to 3.10.0) and Python3.
+The latest version of OpenAeroStruct requires [OpenMDAO](https://github.com/OpenMDAO/openmdao) (versions 3.2.0 to 3.16.0) and Python3.
 Python2 is no longer supported.
 It also requires the folloing packages: `numpy, scipy, matplotlib`.
 If you are looking to use the previous version of OpenAeroStruct which uses OpenMDAO 1.7.4, use OpenAeroStruct 1.0 from [here](https://github.com/mdolab/OpenAeroStruct/releases).
 
 License
 -------
-Copyright 2018-2020 MDO Lab
+Copyright 2018-2022 MDO Lab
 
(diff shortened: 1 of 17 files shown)


eytanadler:
Thanks for the quick update Shugo! I remember Ali went through a similar problem with TACS and [his solution](https://github.com/mdolab/tacs_orig/blob/90fec188483b5cf5770e7c5f651aa6b30d429d6a/tests/test_elements.py#L108) was to set the random seed in the setup method for each test case class like this:

```python
class SampleTestCase(unittest.TestCase):
    def setUp(self):
        np.random.seed(100)
    def test_one(self):
        ...
    def test_two(self):
        ...
    ...
```

I think it might be a cleaner solution since it avoids having to set it in each test case method (the `setUp` method is called before each test case method). What do you think?

kanekosh:
I might be wrong, but if we set a single seed (either using `np.random.seed` or `default_rng`) for each class in the setup method, the random input values generated in each method will depend on the order of execution within each class (i.e., whether the testflo runs `test_one()` first, or `test_two()` first).
I was not confident if the test methods are executed in a deterministic order (especially when parallelizing the test), that's why I hardcoded the test within each method.

eytanadler:
Ali and I were wondering the same thing, but it turns out that `setUp` is run before every test method, not just once for each class. The [documentation](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp) says something vaguely like that, and the example at the bottom of [this page](https://www.codingame.com/playgrounds/10614/python-unit-test-with-unittest) confirms it.

kanekosh:
Thanks Eytan, I didn't know that `setUp` is called every time. I'll fix it soon.

nwu63:
It's explained a bit earlier in the docs:
```
The setUp() and tearDown() methods allow you to define instructions that will be executed before and after each test method.
```

kanekosh:
Addressed Eytan's suggestion

nwu63 (approved):
LGTM

nwu63:
I'm guessing the docs build failure is due to #376 not merged? Should we merge this PR despite the failure?

eytanadler:
Yup, the MDO Lab's RTD is trying to build without actually having the files to tell it which modules to install. Thanks again, Shugo!

PR #376, opened 2022-02-01
Updating to Read-the-docs and Codecov

### LINKED ISSUE #370: Move docs and coverage to RTD and codecov
# Description of feature

Move OpenAeroStruct docs to the MDO Lab's RTD and coverage reporting from coveralls to codecov. This was just done on OpenConcept, so looking at `openconcept.yaml` and `docs/conf.py` in [this commit](https://github.com/mdolab/openconcept/commit/a512c84ed06e0c9c2627d445428bed533ada4557) might help.

lamkina:
## Purpose

This PR updates the OpenAeroStruct docs to ReadtheDocs using the MDO Lab theme.  Additionally, the tests will be run with `pytest` instead of `testflo` to facilitate the switch from coveralls to Codecov.  This is a necessary change to enable the output of a coverage report with an `xml` extension that can be read by Codecov.

Currently, some of the tests fail because of an issue being addressed in this PR: #375.  Once PR #375 is merged, this PR can be merged.

Closes #370.

## Type of change

Maintenance

- [x] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
The github actions build succeeds but it depends on the random seed issue in PR #375 

## Checklist


openaerostruct/docs/conf.py
```diff
@@ -93,29 +92,6 @@
...
-# # The name of an image file (relative to this directory) to place at the top
-# # of the sidebar.
-# html_logo = '_static/OpenMDAO_Logo.png'
-#
-# # The name of an image file (within the static path) to use as favicon of the
-# # docs.  This file should be a Windows icon file (.ico) being 16x16 or 32x32
-# # pixels large.
-# html_favicon = '_static/OpenMDAO_Favicon.ico'
-
 # If not '', a 'Last updated on:' timestamp is inserted at every page bottom,
```
nwu63, line 95: I think some of this can probably be cleaned up and removed

.github/workflows/oas.yml
```diff
@@ -40,22 +39,19 @@ jobs:
...
-        pip install -e ./pygeo
+        pip install "pygeo[testing] @ git+https://github.com/mdolab/pygeo.git"
 
     - name: Run tests
       env:
         OMPI_MCA_btl: ^openib   # prevent OpenMPI warning messages
       run: |
-        cd openaerostruct
-        testflo -n 2 openaerostruct --coverage --coverpkg openaerostruct --cover-omit \*tests/\* --cover-omit \*docs/\*
+        python -m pytest --cov-config=.coveragerc --cov=openaerostruct --cov-report=xml
```
nwu63, line 48: Why the switch to `pytest` from `testflo`?

nwu63 (commented):
See comments below. Mostly look good

.github/workflows/oas.yml
eytanadler: There appears to be no support in testflo for coverage file formats compatible codecov (xml, json, or txt). We were seeing the same problems as I ran into in OpenConcept when I didn't use the --cov-report=xml flag. Maybe there's a solution with testflo given that it's used on other repos? Though you suggested that pytest is much better supported anyway, so we figured it couldn't hurt.

eytanadler:
@nwu63 should we be updating the patch version (to 2.3.3) for changes like these (plus Shugo's)?

.github/workflows/oas.yml
nwu63: We run `coverage` manually to convert the coverage report to xml, see [here](https://github.com/mdolab/.github/blob/master/azure/azure_build.yaml#L112) for the implementation on Azure. If you're not using any testflo features and you don't plan to (MPI testing primarily), then this is fine. I just find it a bit strange given `testflo` is developed by the OM team but no strong objections here.

nwu63:
> @nwu63 should we be updating the patch version (to 2.3.3) for changes like these (plus Shugo's)?

I personally wouldn't since there are no user-facing changes, but there's also no harm in making a new version.

openaerostruct/docs/conf.py
lamkina: Good catch, I'll take that out shortly.

kanekosh:
Thanks Andrew! One thing, `v1_v2_conversion.rst` seems broken a little bit. We used to have side-by-side comparisons of v1 and v2 scripts (by `embed-compare`). I believe `embed-compare` was coming from the OM docs theme.

I personally think we don't have to keep the side-by-side format, but in the current RTD, we cannot tell which script is which version. Could you add a few comments explaining which code block is which version?

.github/workflows/oas.yml
A-CGray: Yeah this is weird, I use testflo in my FEMpy repo and am able to upload to codecov without doing any conversion to xml, not saying you need to figure it out if what you have setup works, but it's definitely possible in general.

lamkina:
> Thanks Andrew! One thing, `v1_v2_conversion.rst` seems broken a little bit. We used to have side-by-side comparisons of v1 and v2 scripts (by `embed-compare`). I believe `embed-compare` was coming from the OM docs theme.
> 
> I personally think we don't have to keep the side-by-side format, but in the current RTD, we cannot tell which script is which version. Could you add a few comments explaining which code block is which version?

I will fix the docs to show which script belongs to v1 and v2.  I'm curious as to why the `embed_n2` extension from the OM docs still works but no the `embed-compare`.  Must be over-ridden by the theme or something.

.github/workflows/oas.yml
nwu63: You seem to be using v1 of the action, where they supported the other format but it's deprecated. They dropped it in v2 --- there have been complaints on their repo since this was an undocumented change.

lamkina: I'm going to commit most of the changes that were brought up shortly and make sure the tests pass using `pytest` after the fix to the random seeds.  If the tests pass, I can change the gh action to manually run `coverage` and revert back to `testflo` to stay consistent with the previous test framework if that seems like the best route.

lamkina:
We found out that the docs reference extensions from unsupported OpenMDAO versions.  Thanks to @kanekosh for tipping us off to this situation.  The plan now is to remove the dependency on OpenMDAO docs extensions.

lamkina:
This PR should be done and ready to merge.  I removed the dependency on old OpenMDAO docs packages and updated the docs to use default sphinx directives.  The docs are now fully moved over the RTD with the MDO Lab theme.

 We found out that OAS was running all the pyspline and pygeo tests due to a directory bug in the GH actions workflow, so I fixed that as well and now all tests are passing thanks to PR #375.  We decided to stick with `pytest` for now, but the switch to testflo seems fairly easy if it becomes necessary in the future.

eytanadler (approved):
Looks good to me, thanks Andrew!

kanekosh (approved):

setup.py
```diff
@@ -7,7 +7,10 @@
     open("openaerostruct/__init__.py").read(),
 )[0]
 
-optional_dependencies = {"docs": ["openmdao[docs]>=3.2, <=3.9.2"], "test": ["testflo>=1.3.6"]}
+optional_dependencies = {
+    "docs": ["openmdao[docs]>=3.2, <=3.9.2", "sphinx_mdolab_theme"],
```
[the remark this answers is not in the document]
eytanadler: On this note, do we even need the `openmdao[docs]` package anymore? As far as I know, there are no more dependencies now that Andrew has moved over the necessary code, but I may have forgotten something

.readthedocs.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+version: 2
+
+python:
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs
+  system_packages: true
```
bernardopacini, line 11: This needs a newline.

.github/workflows/oas.yml
```diff
@@ -90,45 +87,4 @@ jobs:
...
-
-    - name: Push changes to gh-pages
-      if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
-      uses: ad-m/github-push-action@master
-      with:
-        branch: gh-pages
-        directory: gh-pages
-        github_token: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+        black openaerostruct --check --diff -l 120 --target-version py38
```
bernardopacini, line 90: This needs a new line.

bernardopacini (approved):
You all have already discussed this at great length so I don't think I have much to add here. Just two formatting comments to be consistent with Github / the rest of our codes.

.coveragerc
```diff
@@ -2,5 +2,7 @@
 omit =
     */openaerostruct/utils/plot_wing.py
     */openaerostruct/utils/plot_wingbox.py
+    \*tests/\*
```
nwu63, line 5: Are the backslashes here necessary?

openaerostruct/docs/_exts/embed_n2.py
```diff
@@ -0,0 +1,106 @@
+from docutils import nodes
```
nwu63, line 1: Where is this coming from? Are we just copying code from OM?

nwu63 (commented):
Why did we remove the OM docs dependency? I guess it's because they switched to Jupyter stuff right? I quite liked the old system because you're not hard-coding outputs into the docs, so what's shown is always up to date and consistent with the code. For repos in the OM ecosystem, I'm more inclined to do things the same way they're doing it since they may already developed documentation tools etc., but I'm okay with this if Jupyter is not an option/something to consider down the road.

.readthedocs.yaml
lamkina: Good eye

.coveragerc
lamkina: Responding to the comment about the OM docs dependency, we were using sphinx extensions from unsupported OM versions, so we thought it best that for the time being we would transition to sphinx default directives with static outputs.  We are aware this isn't ideal and I think we should discuss the a transition to Jupyter based docs at the maintenance meeting on Monday.

lamkina: As far as the backslashes, that was existing in the `conf.py` file, so I just copied and pasted the regex into the `.coveragerc`.  I'm don't have a great familiarity with regex, so if you think the backslashes should be removed I can get rid of them.

openaerostruct/docs/_exts/embed_n2.py
lamkina: When we removed the OM dependency, this was the only extension we couldn't replace with a default sphinx directive.  Embedding raw html into the docs caused several formatting errors that I couldn't fix given the default sphinx options.  This is the same approach taken by `openconcept`, but I think we should strongly consider moving to Jupyter based docs to eliminate all these issues.

nwu63: I would prefer a comment in the file explaining where this code is from, and perhaps link to the file on GitHub (from the version where this was taken).

setup.py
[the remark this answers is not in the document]
lamkina: This was removed in this [commit](https://github.com/mdolab/OpenAeroStruct/pull/376/commits/1215cc24218f8c636c4dc8f4035d066944dfa512)

openaerostruct/docs/_exts/embed_n2.py
lamkina: I agree that is a good idea.  I will add it now

.coveragerc
nwu63: My guess is that they should mirror the ones above, e.g. `*/openaerostruct/tests/*` etc. After making the changes, please check that `codecov` is correctly skipping these

nwu63 (approved):

PR #392, opened 2022-06-23
Update aerostruct_groups.py

acmarta:
Promote section lift coefficient distribution, Cl, in surfaces

## Type of change

- [x] Other (please describe)

PR #393, opened 2022-06-25
Update __init__.py

acmarta:
Promoted "CDw" and "Cl" in aerostruct group

## Purpose
Makes access to both surface CDw and Cl consistent between aero and aerostruct problems.

## Expected time until merged
Non-urgent PR.

## Type of change
- [x ] Other (please describe)


eytanadler:
Handled in #392

PR #397, opened 2022-08-11
Fixed viscous drag, moment coeff, and sectional CL bugs

### LINKED ISSUE #395: Viscous drag calculation
Hello,

I have a question regarding the viscous drag calculation. The viscous drag of the surface area is calculated via the widths of the panel, which increase with larger sweep backwards. Therefore the drag increases, but the reference surface area between the swept and unswept wing stays the same. I compared this to the OpenVSP results and there the viscous drag remained the same for both wings. Also a CFD simulation showed that the viscous drag remains the same. That is a reason why the optimizer will optimize for a higher sweep (with sweep, chord cp and twist as a design variable) when the induced drag is used as an objective compared to when the total drag is used as the design variable. When increasing the maximum twist angle the optimizer can utilize, the optimized sweep also increases.

kanekosh:
## Purpose
- Fixed bugs in viscous drag, moment coefficient, and sectional CL computation
- Renamed variables name: `cos_sweep` -> `widths`; `widths` -> `lengths_spanwise`.
- Patch version increment

The variable `widths` was used for the length of the quarter chord line for each panel. This is identical to the spanwise width of panels when the sweep is 0, but the `widths` is stretched diagonally when the wing is swept.
Besides, the `cos_sweep` variable was used for the spanwise width, which is confusing.

With this PR, the `widths` variable is now indeed the spanwise width of each panel (which isn't changed by sweep), and the quarter chord line length is now named `lengths_spanwise`.

Because of these confusing namings, we had bugs in viscous drag comp, moment coeff comp, and sectional CL comp. In these components, the computed panel area was larger than the actual area. This PR fixes the bug and closes #395 
The wave drag component was correct, and only the variable names were changed there.

This bugfix changes the drag outputs of all analyses/optimizations with swept wing and all CM outputs. I updated the test reference values accordingly.

## Expected time until merged
A few days

## Type of change
- [x] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

NOTE: I changed the reference values for swept wing cases and moment coefficients. All the other tests passed without updating the ref values.


kanekosh:
Ready for review now

openaerostruct/aerodynamics/viscous_drag.py
```diff
@@ -72,7 +72,7 @@ def compute(self, inputs, outputs):
             S_ref = inputs["S_ref"]
             widths = inputs["widths"]
             lengths = inputs["lengths"]
-            cos_sweep = inputs["cos_sweep"] / widths
+            cos_sweep = inputs["widths"] / inputs["lengths_spanwise"]
```
eytanadler, line 75: Maybe I'm missing something, but I no longer see the bug here? Won't this have the same value as before?

eytanadler (commented):
Thanks for the update Shugo! Not sure what I saw before, but I don't see the bug in the viscous drag; am I looking in the wrong place? It seems to me that the geometry stuff is staying the same, with new names, and then the new names are replaced in the calculations. Can you point me to where the bug was that I'm missing?

openaerostruct/aerodynamics/lift_coeff_2D.py
```diff
@@ -45,7 +45,7 @@ def setup(self):
         # Inputs
         self.add_input("alpha", val=3.0, units="deg")
         self.add_input("sec_forces", val=np.ones((self.nx - 1, self.ny - 1, 3)), units="N")
-        self.add_input("widths", val=np.arange((self.ny - 1)) + 1.0, units="m")
+        self.add_input("widths", val=np.ones((self.ny - 1)) * 0.2, units="m")
```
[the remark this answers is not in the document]
kanekosh: I needed to set the same default value for the "widths" variable across several different components, otherwise OM raises an error

openaerostruct/aerodynamics/viscous_drag.py
kanekosh: The bug in viscous drag computation was coming from the `widths` variable, which was the diagonally stretched spanwise width. The viscous drag component had been using `widths` to compute the panel area, hence the overestimation of viscous drag.
In other words, the bug was because of the inconsistency of the `widths` definition between the viscous drag component and geometry component (so we could say the bug was not in viscous drag itself, but it was about geometry variable naming).

Now I modified the OM variable names, and the `widths` variable is now actually the width along the y-axis (or zero sweep direction). The actual bug fix was done in `aerodynamics/geometry.py` where I changed the definition of the `widths` variable.

And yes, the `cos_sweep` python variable has the same value as before.

The same applies to the CM and sectional CL fixes; those bugs were fixed by changing the definition of `widths` in the geometry component. That's why you don't see the code changes in the CM and sectional CL components.

openaerostruct/aerodynamics/geometry.py
```diff
@@ -22,6 +22,10 @@ class VLMGeometry(om.ExplicitComponent):
         Bound points for the horseshoe vortices, found along the 1/4 chord.
     widths[ny-1] : numpy array
         The spanwise widths of each individual panel.
```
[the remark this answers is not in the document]
kanekosh: Good point, I updated the explanation here

openaerostruct/aerodynamics/viscous_drag.py
eytanadler: Ah I see it now, thanks for the help. It was a bit sneaky that by changing the variable name it fixed something later that you didn't edit 😏

eytanadler (approved):

PR #400, opened 2023-02-11
Allow newest version of OpenMDAO

eytanadler:
## Purpose

Removed the upper bound on the OpenMDAO version. Because there may be changes in new OpenMDAO versions (or other packages for that matter) that break parts of OpenAeroStruct, I also added automatic GitHub Actions testing twice a month. This will tell us if any changes in the newest package versions have caused errors in OpenAeroStruct.

Do we want to update the lower bound on OpenMDAO? Version 3.2.0 is getting pretty old. We aren't testing intermediate versions of packages so we don't really know that older versions still work.

## Expected time until merged

Within a couple weeks?

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/openaerostruct/docs/installation.rst
+++ b/openaerostruct/docs/installation.rst
@@ -26,7 +26,7 @@ Then from within the OpenAeroStruct folder, pip install the package:
 
 Both methods will automatically install the dependencies: numpy, scipy, matplotlib, and OpenMDAO.
 
-The latest version of OpenAeroStruct supports OpenMDAO version 3.2.0 to 3.16.0; 3.16.0 is recommended.
+The latest version of OpenAeroStruct supports OpenMDAO versions 3.2.0 or newer; the newest is recommended as it is tested regularly.
 If you wish to install a specific version of OpenMDAO, follow the instructions at https://github.com/OpenMDAO/OpenMDAO/.
 
 If you are unfamiliar with OpenMDAO and wish to modify the internals of OpenAeroStruct, you should examine the OpenMDAO documentation at http://openmdao.org/twodocs/versions/latest/index.html. The tutorials provided with OpenMDAO are helpful to understand the basics of using OpenMDAO to solve an optimization problem.
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ setup(
     # Test files
     package_data={"openaerostruct": ["tests/*.py", "*/tests/*.py", "*/*/tests/*.py"]},
     install_requires=[
-        "openmdao>=3.2, <=3.16.0",  # don't forget to update OM versions in README.md and docs/installation.rst
+        "openmdao>=3.2",
         "numpy>1.19",
         "scipy>=1.6.0",
         "matplotlib",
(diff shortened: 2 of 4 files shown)


kanekosh:
Thanks @eytanadler , looks good to me!
I think we should also update the OM lower bound. Do we also want to test the oldest OM version in the Github Action by defining a matrix? E.g. Dymos does that: https://github.com/OpenMDAO/dymos/blob/master/.github/workflows/dymos_tests_workflow.yml

I personally wouldn't bother testing it because it doubles the build time, but I'm okay either way.

eytanadler:
> Thanks @eytanadler , looks good to me! I think we should also update the OM lower bound. Do we also want to test the oldest OM version in the Github Action by defining a matrix? E.g. Dymos does that: https://github.com/OpenMDAO/dymos/blob/master/.github/workflows/dymos_tests_workflow.yml
> 
> I personally wouldn't bother testing it because it doubles the build time, but I'm okay either way.

I'll add a test for older OM versions. I believe GHA runs the build matrix in parallel, so it shouldn't slow it down much. On second thought, I guess if the version works there's no point in updating it since leaving as is will enable pip to resolve versions more easily if there are dependency conflicts.

eytanadler:
@kanekosh @bernardopacini I think this is finally ready for review. The GHA build file isn't the prettiest thing you'll ever see, but it does the job it needs to.

setup.py
```diff
@@ -11,7 +11,7 @@
 optional_dependencies = {
     "docs": ["sphinx_mdolab_theme"],
     "test": ["pytest", "pytest-cov", "coverage"],
-    "ffd": ["pygeo>=1.6.0"]
+    "ffd": ["pygeo>=1.6.0"],
```
kanekosh, line 14: pygeo>=1.12.2 ?

kanekosh (changes requested):
Thank you Eytan! Other than the pygeo version in setup.py, looks good to me

setup.py
eytanadler: Good catch. pyGeo 1.12.2 was the latest that works with the old versions of the other packages, but older versions of pyGeo work too. I'll see if 1.6.0 works with the old build and if so leave that as the old build version.

kanekosh (approved):

openaerostruct/docs/installation.rst
```diff
@@ -26,20 +26,51 @@ Then from within the OpenAeroStruct folder, pip install the package:
...
+      - latest
+    * - Matplotlib
+      - latest
+      - latest
+    * - pyGeo (optional)
+      - 1.6.0
+      - latest
+    * - OpenVSP (optional)
+      - 3.27.1
+      - 3.27.1
```
bernardopacini, line 57: Is this duplicate by mistake?

bernardopacini (approved):
One minor comment, but overall looks good to me!

openaerostruct/docs/installation.rst
eytanadler: This is intentional since the OpenVSP version is 3.27.1 on both the oldest and latest GitHub Actions builds

PR #410, opened 2023-05-09
Add the possibility to change the position of the scaling for the chord design variable

lucaeros:
## Purpose

Scaling position at quarter chord :
Screenshot from 2023-05-09 11-40-05

Scaling position at trailing edge :
Screenshot from 2023-05-09 11-51-05

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


kanekosh:
Thanks Lucas! Could you add a quick documentation about this new feature in the [geometry manipulation](https://mdolab-openaerostruct.readthedocs-hosted.com/en/latest/advanced_features/geometry_manipulation.html) and [surface dict entries](https://mdolab-openaerostruct.readthedocs-hosted.com/en/latest/user_reference/mesh_surface_dict.html#mesh-dict) pages?

openaerostruct/geometry/geometry_mesh_transformations.py
```diff
@@ -131,11 +131,16 @@ def initialize(self):
         """
         self.options.declare("val", desc="Initial value for chord lengths")
         self.options.declare("mesh_shape", desc="Tuple containing mesh shape (nx, ny).")
+        self.options.declare(
+            "gamma",
```
eytanadler, line 135: Could you pick a more intuitive name then `gamma` for the option? Maybe keeping the same as the name in the surface dictionary would make sense

openaerostruct/geometry/geometry_mesh.py
```diff
@@ -79,10 +79,15 @@ def setup(self):
         val = np.ones(ny)
         if "chord_cp" in surface:
             promotes = ["chord"]
+            if "chord_scaling_pos" in surface:
+                gamma = surface["chord_scaling_pos"]
```
eytanadler, line 83: Once you change the option name, also change this variable name. Also can't this if else block be moved outside of the `if "chord_cp" in surface` block so that you don't need to duplicate the `gamma = 0.25` line? If you want to be even nicer, you could throw a warning if `"chord_scaling_pos"` is specified but there are no chord DVs.

openaerostruct/geometry/geometry_mesh_transformations.py
```diff
@@ -131,11 +131,16 @@ def initialize(self):
...
+            default=0.25,
+            desc="float in [0.,1.] which indicates the chord fraction where to do the chord scaling. 1. is trailing edge, 0. is leading edge",
```
eytanadler, line 137: Could it still be valid if this value is not between 0 and 1? If not, add a check in `setup` that throws an error if gamma is < 0 or > 1

eytanadler (changes requested):
Thanks for the contributions! In addition to the code comments I added, please also do the following:

1. Add this new feature to the rectangular wing chord optimization example [here](https://github.com/mdolab/OpenAeroStruct/blob/dcdb76bf931f881f05f6c39c74024b5b305912f4/openaerostruct/examples/rectangular_wing/opt_chord.py#L60).
2. Add a `"chord_scaling_pos"` entry to the surface dictionary in [this test](https://github.com/mdolab/OpenAeroStruct/blob/main/openaerostruct/geometry/tests/test_geometry_mesh.py) to test that your change works.
3. Add another test for `ScaleX` in [this file](https://github.com/mdolab/OpenAeroStruct/blob/main/openaerostruct/geometry/tests/test_geometry_mesh_transformations.py) that checks the partials and confirms that the chord values have changed as you expect.

openaerostruct/geometry/geometry_mesh_transformations.py
lucaeros: Works for any float values

openaerostruct/geometry/geometry_mesh.py
```diff
@@ -77,12 +77,21 @@ def setup(self):
...
 
         val = np.ones(ny)
+        chord_scaling_pos = 0.25  # if no scaling position is specified : chord scaling w.r.t quarter of chord
         if "chord_cp" in surface:
             promotes = ["chord"]
+            if "chord_scaling_pos" in surface:
+                chord_scaling_pos = surface["chord_scaling_pos"]
         else:
+            if "chord_scaling_pos" in surface:
+                print("WARNING: chord_scaling_pos has been specified but no chord design variable available")
```
eytanadler, line 87: Use the built in warnings package as we do [here](https://github.com/mdolab/openconcept/blob/c92037ed8a03c29030c6ea8c22357fbb279f95ef/openconcept/utilities/math/integrals.py#L602). It's nicer because it allows the user to turn off warnings if desired (among other things)

openaerostruct/geometry/tests/test_geometry_mesh_transformations.py
```diff
@@ -132,6 +138,28 @@ def test_scalex_symmetry(self):
...
 
+    def test_scalex_chord_scaling_pos(self):
+        symmetry = False
+        mesh = get_mesh(symmetry)
+
+        prob = om.Problem()
+        group = prob.model
+
+        val = self.rng.random(NY)
+        chord_scaling_pos = self.rng.random(1)
```
eytanadler, line 149: Can you pick a value here, maybe 0 or 1, that you can analytically check? Then add an assert after run model that checks that the chord has been modified as you expect

eytanadler (changes requested):
Thanks for the updates! Other than the two comments, please add documentation to [geometry manipulation](https://mdolab-openaerostruct.readthedocs-hosted.com/en/latest/advanced_features/geometry_manipulation.html) as Shugo mentioned about this new feature. It'd be nice to add two new pictures in the same format as the others to show what it does as there are for the existing design variables.

lucaeros:
> Thanks for the updates! Other than the two comments, please add documentation to [geometry manipulation](https://mdolab-openaerostruct.readthedocs-hosted.com/en/latest/advanced_features/geometry_manipulation.html) as Shugo mentioned about this new feature. It'd be nice to add two new pictures in the same format as the others to show what it does as there are for the existing design variables.
I added the feature here :
https://mdolab-openaerostruct.readthedocs-hosted.com/en/latest/user_reference/mesh_surface_dict.html#mesh-dict

eytanadler:
Last thing, this one is easy. Could you update the version in the openaerostruct/__init__.py file to 2.6.2? Then I think this is ready

eytanadler (approved):

openaerostruct/geometry/tests/test_geometry_mesh_transformations.py
bernardopacini: Is this still an open item?

bernardopacini (approved):
This looks good to me, but what about the open item @eytanadler mentioned above? I agree that fixing it to a value and checking the result is a good idea.

lucaeros:
> This looks good to me, but what about the open item @eytanadler mentioned above? I agree that fixing it to a value and checking the result is a good idea.

A test has been added to check if the` chord_scaling_pos` is set to 1, the Trailing edge should not move when using chord DV.

lucaeros:
> > This looks good to me, but what about the open item @eytanadler mentioned above? I agree that fixing it to a value and checking the result is a good idea.
> 
> A test has been added to check if the` chord_scaling_pos` is set to 1, the Trailing edge should not move when using chord DV.

Otherwise, the derivatives are check for any number of `chord_scaling_pos`

bernardopacini (approved):
Thanks for adding the test, that looks good to me.

eytanadler:
One more change, but I can make this one. The new test @lucaeros added should be in its own function, not within an existing test method.

eytanadler (approved):

PR #412, opened 2023-06-19
Fixing oldest OpenMDAO install

kanekosh:
## Purpose

This PR fixed the build with the oldest dependencies (specifically, the installation of OpenMDAO 3.15.0)

## Expected time until merged

Soon

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/workflows/oas.yml
+++ b/.github/workflows/oas.yml
@@ -25,7 +25,7 @@ jobs:
         # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Set versions to test here ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
         PYTHON_VERSION_OLDEST: ['3.8']
         PYTHON_VERSION_LATEST: ['3.11']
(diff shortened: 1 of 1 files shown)


eytanadler:
Does the old wheel version automatically use an older pip? Why is the old pip version no longer necessary?

kanekosh:
Since I dropped `--upgrade` the pip is not upgraded to the latest. On my local machine, the build also worked with the newer pip (23.1.2), but I don't think we need the latest pip for the oldest build.

The pip version no longer matters here because with the older wheel, we can now build the wheel for OpenMDAO==3.15.0, thus no need to revert to the legacy `setup.py install` (that was the reason I pinned to the old pip version previously).

eytanadler:
> Since I dropped `--upgrade` the pip is not upgraded to the latest. On my local machine, the build also worked with the newer pip (23.1.2), but I don't think we need the latest pip for the oldest build.
> 
> The pip version no longer matters here because with the older wheel, we can now build the wheel for OpenMDAO==3.15.0, thus no need to revert to the legacy `setup.py install` (that was the reason I pinned to the old pip version previously).

Makes sense, good find. I'm going to give this a shot on https://github.com/mdolab/openconcept/pull/53 too.

eytanadler (approved):

PR #417, opened 2023-09-17
Added dict key validation & updated documentation

kanekosh:
### Key validation
With this PR, OAS will check a user-defined dicts (mesh dict or surface dict) and warn if the user provides a key that is not implemented or used in OAS.

For example, some examples and tests provide the `chord` key when creating a mesh, but the key `chord` does not exist and therefore never used. I fixed them to `root_chord`, the one actually used.
I also removed unused keys from other tests.

### Documentation update
The current documentation and docstrings say the `chord_cp` variable is the actual chord value in meters, but what the code (ScaleX component) actually does is that it applies the chord scaler to the initial mesh. Therefore `chord_cp` is actually a dimensionless chord scaler.  
This PR fixes the documentation and docstrings.

Also, I updated the surface dict documentation.

(It's a little too late but I probably should've opened two separate PRs for this...)

## Type of change

- [x] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing

I added a new test to check the chord values before and after `ScaleX` is applied.

## Checklist

### DIFF
--- a/openaerostruct/utils/check_surface_dict.py
+++ b/openaerostruct/utils/check_surface_dict.py
@@ -12,7 +12,7 @@ def check_surface_dict_keys(surface):
         User-defined surface dict
     """
 
-    # NOTE: make sure this is consistend to the documentation's surface dict page
+    # NOTE: make sure this is consistent to the documentation's surface dict page
     keys_implemented = [
         # wing definition
         "name",
--- a/openaerostruct/utils/check_surface_dict.py
+++ b/openaerostruct/utils/check_surface_dict.py
@@ -1,0 +1,77 @@
+import warnings
+
+
+def check_surface_dict_keys(surface):
+    """
+    Key valication function for the OAS surface dict.
+    Shows a warning if a user provided a key that is (likely) not implemented in OAS.
+
+    Parameters
+    ----------
+    surface : dict
+        User-defined surface dict
+    """
+
+    # NOTE: make sure this is consistend to the documentation's surface dict page
+    keys_implemented = [
+        # wing definition
+        "name",
+        "symmetry",
+        "S_ref_type",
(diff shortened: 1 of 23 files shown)


eytanadler (dismissed):
Thanks for fixing these!

nwu63:
> In addition, I found that some examples and tests provide the chord key when creating a mesh, but the key chord does not exist and therefore never used. I fixed them to root_chord, the one actually used.

Should there be some sort of key validation then? To ensure we aren't passing in dict entries that are silently inactive?

kanekosh:
> > In addition, I found that some examples and tests provide the chord key when creating a mesh, but the key chord does not exist and therefore never used. I fixed them to root_chord, the one actually used.
> 
> Should there be some sort of key validation then? To ensure we aren't passing in dict entries that are silently inactive?

That's true, I'll add it.

kanekosh:
Added key validation as @nwu63 suggested, ready for the 2nd review

openaerostruct/utils/check_surface_dict.py
```diff
@@ -0,0 +1,78 @@
...
+    Key valication function for the OAS surface dict.
+    Shows a warning if a user provided a key that is (likely) not implemented in OAS.
+
+    Parameters
+    ----------
+    surface : dict
+        User-defined surface dict
+    """
+
+    # NOTE: make sure this is consistend to the documentation's surface dict page
```
eytanadler, line 15: Consistend -> consistent

eytanadler (dismissed):
Thanks for doing this tedious work. Only one very minor spelling correction, and otherwise looks good!

eytanadler (approved):

kanekosh:
Thanks for the quick review!

Issue #423, opened 2024-02-26
The method for adjusting zoom in plot_wingbox.py does not work with latest matplotlib


shamsheersc19:
## Description
In plot_wingbox.py, `ax.dist` is used to adjust the zoom for the wing visualization. This attribute and method have been in deprecated in matplotlib. For example, it doesn't work for v3.8. The correct way to do this for recent version is as described in https://py4u.org/questions/75146691/

### Steps to reproduce issue

1. Get a recent matplotlib version e.g., 3.8.3
2. Run the run_aerostruct_uCRM_multipoint.py example
3. Call the plot_wingbox.py script as showing the wingbox walkthrough.

### Current behavior
```
Exception in Tkinter callback
Traceback (most recent call last):
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.2288.0_x64__qbz5n2kfra8p0\Lib\tkinter\__init__.py", line 1967, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "C:\Users\shams\OpenMDAO\OpenAeroStruct\openaerostruct\utils\plot_wingbox.py", line 798, in update_graphs
    self.plot_wing()
  File "C:\Users\shams\OpenMDAO\OpenAeroStruct\openaerostruct\utils\plot_wingbox.py", line 593, in plot_wing
    dist = self.ax.dist
           ^^^^^^^^^^^^
AttributeError: 'Axes3D' object has no attribute 'dist'
```

### Expected behavior
The results should be visualized as shown in the wingbox walkthough.

PR #425, opened 2024-03-07
Fix plot_wing.py to work for matplotlib v3.8+

shamsheersc19:
## Purpose
Fix plot_wing.py to work for matplotlib v3.8+.
The `dist` attribute for 3-D plots has been deprecated in recent matplotlib versions.
See related Issue #423

Also made the 3-D view behavior consistent in plot_wingbox.py (the 3-D view should not reset if the iteration is changed).

## Expected time until merged
1 day
Code review should only take a few minutes.

## Type of change
 Bugfix (non-breaking change which fixes an issue)

## Testing
Ran the run_CRM.py and run_aerostruct_uCRM_multipoint.py examples, and also confirmed that the views don't reset it the iteration is changed.
Tested with matplotlib versions:
* 3.6.0
* 3.7.0
* 3.8.0
* 3.8.3

image

image

## Checklist

### DIFF
--- a/openaerostruct/utils/plot_wing.py
+++ b/openaerostruct/utils/plot_wing.py
@@ -465,7 +465,6 @@ class Display(object):
         self.ax.cla()
         az = self.ax.azim
         el = self.ax.elev
-        dist = self.ax.dist
 
         for j, name in enumerate(self.names):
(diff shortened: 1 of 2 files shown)


kanekosh (approved):
Thanks again, @shamsheersc19

PR #428, opened 2024-05-17
Updating MPhys wrapper for 2.0 update

timryanb:
## Purpose
Refactored builder class to use new mphys import procedures and variable naming convention. MPhys 2.0 has officially been released, so this PR is ready to be merged in. This is a backwards-incompatible change for the MPhys wrapper, so I'd recommend making a version bump after this PR. 

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


timryanb:
It looks like the CI tests are failing due to the fact that MPhys 1.1.0 is pinned in the Azure pipeline. Is there a way to update this when MPhys 2.0 is released @A-CGray?

A-CGray:
> It looks like the CI tests are failing due to the fact that MPhys 1.1.0 is pinned in the Azure pipeline. Is there a way to update this when MPhys 2.0 is released @A-CGray?

Ah yeah, this is kind of a pain, we probably won't update the docker images that the CI jobs use to MPhys 2 until our other codes like ADflow and pyGeo are MPhys 2 compatible. In the meantime, I've added a line to the script that install OpenAeroStruct in the CI jobs that removes the pip constraint on the mphys version so that it can install mphys 2 itself.

kanekosh (changes requested):
Thank you, Tim!
I believe `integration_tests/test_simple_rect_mphys_aero.py` and `test_simple_rect_mphys_aero_compressible.py` also need to be updated. These tests are currently skipped (rather than tests being failed) as a result of failed Mphys import.

timryanb:
> Thank you, Tim! I believe `integration_tests/test_simple_rect_mphys_aero.py` and `test_simple_rect_mphys_aero_compressible.py` also need to be updated. These tests are currently skipped (rather than tests being failed) as a result of failed Mphys import.

Good catch @kanekosh, should be fixed now

PR #430, opened 2024-06-03
Minor fix to mesh generator

### LINKED ISSUE #378: Mesh `num_x` restriction
## Description
`num_x` of a mesh should be able to be any number (even or odd). However, the current code requires `num_x` to be 2 or an odd number (it cannot be 4, 6, 8, ...), which doesn't make sense. See:
https://github.com/mdolab/OpenAeroStruct/blob/master/openaerostruct/geometry/utils.py#L649

### TODO
- See if I can remove that restriction and the code works with `num_x = 4`
- See if the output makes sense physically (do "mesh refinement study" to see the trend)

kanekosh:
## Purpose
OAS mesh generator currently only accepts the chordwise discretization `nx` to be 2 or an odd number because of how it generates the cosine spacing in the chordwise direction. This PR fixes the mesh generator to make it work with even numbers for `nx`.

I also realized a weird behavior for rectangle wings: the default chordwise coordinate for [LE, TE] was `[0, chord]` when `nx=2` and `[-chord/2, chord/2]` when `nx >= 3`. Now I fixed it, and the default coordinate is `[0, chord]` for any `nx`. I adjusted a few test cases accordingly because it affects CM computation.

Also made a minor edit to the GHA setting so that the "latest" build continues even if the "oldest" job fails, and vice versa.

Closes #378 

### Mesh examples
Screenshot 2024-06-03 at 12 06 58 PM
Screenshot 2024-06-03 at 12 07 12 PM
Screenshot 2024-06-03 at 12 45 52 PM

## Expected time until merged

Week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


kanekosh:
Codecov is failing, but it's ready for review.

openaerostruct/geometry/utils.py
```diff
@@ -396,23 +397,18 @@ def gen_rect_mesh(num_x, num_y, span, chord, span_cos_spacing=0.0, chord_cos_spa
...
-
-    # Special case if there are only 2 chordwise nodes
-    if num_x <= 2:
-        full_wing_x = np.array([0.0, chord])
+    # --- chordwise discretization ---
+    cosine = -0.5 + 0.5 * (1 - np.cos(np.linspace(0, np.pi, num_x)))  # cosine spacing from -0.5 to 0.5
+    uniform = np.linspace(-0.5, 0.5, num_x)  # uniform spacing
+    # mixed spacing with chord_cos_spacing as a weighting factor
+    wing_x = cosine * chord_cos_spacing + (1 - chord_cos_spacing) * uniform  # from -0.5 to 0.5
+    wing_x += 0.5  # offset by 0.5 so it goes 0 to 1 (leading edge = 0, trailing edge = 1)
```
eytanadler, line 405: Minor, but why include this offset by 0.5 line after already shifting the cosine and uniform distributions by -0.5? I think may as well just have them both go from 0 to 1 and get rid of this line. As far as I can tell, they should give the same result.

openaerostruct/geometry/utils.py
```diff
@@ -557,24 +553,13 @@ def add_chordwise_panels(mesh, num_x, chord_cos_spacing):
...
-
-    else:
-        # Mirror this half wing into a full wing; offset by 0.5 so it goes 0 to 1
-        full_wing_x = np.hstack((-half_wing[:-1], half_wing[::-1])) + 0.5
+    # chordwise discretization
+    cosine = -0.5 + 0.5 * (1 - np.cos(np.linspace(0, np.pi, num_x)))  # cosine spacing from -0.5 to 0.5
+    uniform = np.linspace(-0.5, 0.5, num_x)  # uniform spacing
+    # mixed spacing with chord_cos_spacing as a weighting factor
+    wing_x = cosine * chord_cos_spacing + (1 - chord_cos_spacing) * uniform  # from -0.5 to 0.5
+    wing_x += 0.5  # offset by 0.5 so it goes 0 to 1
```
eytanadler, line 562: Same comment as above

eytanadler (changes requested):
This makes the code both simpler and more general, very nice! (Nevermind about bumping the patch, I got it)

If you don't mind,  I have a couple things to try with GHA to try to get codecov working that I'll add here. No guarantees that it'll work though.

openaerostruct/geometry/utils.py
kanekosh: Good catch, fixed

openaerostruct/geometry/utils.py
kanekosh: fixed

sabakhshi (approved):
These changes make sense. The inconsistent chordwise positioning of the wing added extra complexity when doing stability analysis. These changes will break my 740 code as result but they are for the better.

kanekosh:
@sabakhshi I'm planning on adding a stability-constrained example to the docs, so I'll document the origin of the chordwise coordinate at that time.

eytanadler (approved):

PR #434, opened 2024-07-15
Run github actions on mdolab docker images

A-CGray:
## Purpose

The current github action for running the OAS tests is a bit of a mess, it tries to build OpenMPI, PETSc and OpenVSP, and it uses "latest" and "old" package versions that don't match the rest of our packages. It would be much cleaner if we could run the OAS CI jobs on our docker containers where we have working environments with all the dependencies OAS needs and with stable and latests versions of things that match our other packages.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


A-CGray:
For reasons I'm clearly to stupid to understand, GitHub actions can only run in a docker container if they run as the root user (see https://github.com/actions/checkout/issues/1014 and https://github.com/actions/checkout/issues/956). This doesn't work for our docker containers because we install things under the `mdolabuser` account.

The only option I see if we want to run OAS's CI on our docker images is to migrate it to using Azure. Any other ideas @eirikurj ?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,16 @@
+name: OAS release
```
eirikurj, line 1: Is there a need for a separate GHA file? I think it would make sense to add this in as a stage in the azure script. That way you can add a dependency, i.e., only push/publish if builds/tests stages pass.

eirikurj (commented):
Yes, I agree that moving testing to our docker containers is the way to go. I can add an Azure pipeline, but it will need to depend on this branch if you want to test it, and once merged we need to change it to main. Let me know what you prefer.

A-CGray:
> I can add an Azure pipeline, but it will need to depend on this branch if you want to test it, and once merged we need to change it to main. Let me know what you prefer.

Will that work given that this PR is from my fork? If so then yes let's do that to make sure it works

.github/workflows/release.yaml
A-CGray: All our other pypi codes use this GHA as I don't believe any of them use Azure, I think it's fine to not have this depend on tests passing etc because it's only run when we manually create a new release

ewu63: We do use Azure for e.g. baseclasses [here](https://github.com/mdolab/baseclasses/blob/main/.github/azure-pipelines.yaml#L25) and the sphinx theme. You're right that most of the repos use GHA though (and are completely separate in CI toolchain from Azure).

A-CGray: Oh right, I didn't realise that, I'll follow the baseclasses example then

eirikurj:
> Will that work given that this PR is from my fork? If so then yes let's do that to make sure it works

No, the branch needs to be on under the mdolab account to have access to it.

A-CGray:
> > Will that work given that this PR is from my fork? If so then yes let's do that to make sure it works
> 
> No, the branch needs to be on under the mdolab account to have access to it.

OK, are you able to make a copy of this branch on the mdolab fork then? I don't have the permissions to do it

eirikurj:
Hmm, strange MDO Lab member only have triage permissions, but we typically have write for member. Not sure why, but we can discuss this during next maintenance meeting. New branch is here https://github.com/mdolab/OpenAeroStruct/tree/GHAImprovement

A-CGray:
Thanks, closing now as this is superseded by https://github.com/mdolab/OpenAeroStruct/pull/438

PR #435, opened 2024-07-17
Make geometry spline interpolation mesh independent

A-CGray:
### Problem
OpenAeroStruct uses b-splines to interpolate geometric design variable values onto the mesh (either at mesh nodes or panel mid-points), these b-splines are defined over a coordinate range of 0 to 1, so it is necessary to map the spanwise coordinates of the mesh to the same range. In other words, we need to define where on the spline each mesh point interpolates it's design variable value from.

Currently, for all DVs that use this interpolation, we simply linearly space these interpolation points from 0 to 1. This works fine if your mesh has constant spanwise spacing but otherwise produces unintuitive results, it also means that for the same design variable values, changing the spanwise mesh distribution significantly changes the wing geometry.

Additionally, this approach means that for the t_over_c variable, the b-spline used for interpolation technically spans not from the root to the tip but from the centre of the first panel to the centre of the last panel. This means that, even if you keep a constant spanwise mesh spacing, the interpolation of t_over_c values will change depending on how many panels you use along the span.

### Fix

This PR fixes the above issue by computing the points on the b-spline to interpolate from based on the actual mesh coordinates. This means that the wing geometry no longer depends on the spanwise mesh distribution, and the t_over_c interpolation no longer depends on the number of spanwise mesh points.

### Example

In the case below I use the standard rectangular wing mesh and control x-shear and chord with 2 control points, one at the tip and one at the root. This should result in the chord length and x-shear being linearly interpolated on the wing, but with the current implementation this is not the case if you use cosine spanwise spacing (shown in blue). The Orange mesh is the geometry produced with the changes in this PR and is as you'd expect.

image

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

A few of the tests currently fail because this PR slightly alters how t_over_c is interpolated which slightly changes the optimized reference values we test against, if you're happy with these changes I will update the reference values.

## Checklist

### DIFF
--- a/openaerostruct/geometry/geometry_group.py
+++ b/openaerostruct/geometry/geometry_group.py
@@ -2,34 +2,7 @@ import numpy as np
 
 import openmdao.api as om
 from openaerostruct.utils.check_surface_dict import check_surface_dict_keys
-
-
-def get_normalized_span_coords(surface, mid_panel=False):
-    """Get the normalised coordinates used for interpolating values along the wingspan
-
-    These normalized coordinates range from 0 at the tip of the wing to 1 at the root.
-
-    Parameters
-    ----------
-    surface : OpenAeroStruct surface dictionary
-        Surface to generate coordinates for
-    mid_panel : bool, optional
-        Whether the normalized coordinate should be of the panel midpoints rather than the mesh nodes, by default False
-
-    Returns
-    -------
-    np.array
-        Normalized coordinate values
-    """
-    spanwise_coord = surface["mesh"][0, :, 1]
-    span_range = spanwise_coord[-1] - spanwise_coord[0]
-    span_offset = spanwise_coord[0]
-    if mid_panel:
-        x_real = (spanwise_coord[:-1] + spanwise_coord[1:]) / 2
-    else:
-        x_real = spanwise_coord
-    x_norm = (x_real - span_offset) / span_range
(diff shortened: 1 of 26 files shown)


sabakhshi (commented):
Long standing bug fixed. Changes are minimal for the Geometry group but the tests still need to be updated as mentioned.

kanekosh:
This is great. I remember facing this bug a while ago and ended up using a custom geometry component for my project (bad practice, sorry!). Yes please update the test reference values.

A-CGray:
> This is great. I remember facing this bug a while ago and ended up using a custom geometry component for my project (bad practice, sorry!). Yes please update the test reference values.

Done, I realised there were some other places using B-spline components with the same issue so fixed those too

kanekosh (approved):

A-CGray:
@sabakhshi probably better if you review this, not sure @lamkina is a particularly prolific OAS user

sabakhshi (approved):
Same as the my previous comments. I approve.

lamkina:
I removed myself as a reviewer, but my overall comment is the B-Spline corrections make sense so as long as the tests are passing this should be all set.

A-CGray:
@sabakhshi or @kanekosh can one of you merge this?

PR #438, opened 2024-07-19
Move CI to Azure

A-CGray:
## Purpose

The current github action for running the OAS tests is a bit of a mess, it tries to build OpenMPI, PETSc and OpenVSP, and it uses "latest" and "old" package versions that don't match the rest of our packages. It would be much cleaner if we could run the OAS CI jobs on our docker containers where we have working environments with all the dependencies OAS needs and with stable and latests versions of things that match our other packages.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist


A-CGray:
The VSP test is failing because the vsp file it depends on isn't moved to the python installation directory but the tests are because they're inside the openaerostruct package directory. I think we should move all the tests out of the package directory into their own top level directory like we do for all our other packages

kanekosh (approved):

PR #446, opened 2024-09-23
Implemented multi-section surface parameterization

sabakhshi:
## Purpose
In support of the Army/CDI project I have implemented a multi-section geometry group for OpenAeroStruct that allows the optimizer to have full control over individual wing sections using all existing OpenAeroStruct transformations. Two approaches for facilitating C0 continuity between sections, by constraint or by construction, are provided. In addition, an automatic mesh generator for multi-section wing is implemented. This feature can be back-ported to single-section surfaces in a future PR.

## Expected time until merged
3 weeks

## Type of change

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Tests for multi-section geometries have been added.

## Checklist


kanekosh (changes requested):
Thanks Safa, I'll take a detailed look once you fix the docs build.
Meanwhile, could you add a test that compares analysis results of a single-section geometry and multi-section geometry for the identical wing? E.g. create a simple tapered wing using both multi-section and single-section components, and check that both give the identical CL, CD, etc.

sabakhshi:
Will add the test. I'm wondering what you mean about the docs build failing? Github tells me that is has succeeded.

kanekosh:
Sorry never mind the docs are fine, flake8/black is failing though

sabakhshi:
Ready for review

tests/multisection_integration_tests/test_multi_sym_2_sec_constr.py
```diff
@@ -0,0 +1,147 @@
...
+        prob.driver.options["maxiter"] = 1000
+        prob.driver.options["debug_print"] = ["nl_cons", "objs", "desvars"]
+
+        # Set up and run the optimization problem
+        prob.setup()
+        # prob.run_model()
+        prob.run_driver()
+        # om.n2(prob)
+
+        assert_near_equal(prob["aero_point_0.surface_perf.CD"][0], 0.02919978, 1e-6)
```
kanekosh, line 141: Should the reference value be identical to the one in `test_multi_sym_2_sec.py` as these two look like running the same case with a different way of handling multi-section continuity? Or are they different in setup?

kanekosh (changes requested):
Thank you Safa. In addition to the comments below,
- Please update docs/user_reference/mesh_surface_dict.rst to reflect your additions for multi-section geometry.
- I recommend renaming functions and variables following [PEP-8](https://peps.python.org/pep-0008/). For example, `generateMesh` -> `generate_mesh`.

openaerostruct/aerodynamics/aero_groups.py
```diff
@@ -30,6 +30,37 @@ def setup(self):
...
+                    "CD0",
+                    "with_viscous",
+                    "with_wave",
+                    "groundplane",
+                    "k_lam",
+                    "t_over_c_cp",
+                    "c_max_t",
+                ]
+
+                # Constructs a surface dictionary and adds the specified supported keys and values from the mult-section surface dictionary.
```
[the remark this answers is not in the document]
sabakhshi: It doesn't eliminate them from the original `surface` dict. The issue is that the the multisurface dict is constructed a bit differently than the single surface dict. Giving the multisurface dict to AeroPoint caused issues so what I do is rather than modify AeroPoint, I give AeroPoint a version of the multisurface dict that only contains information necessary for AeroPoint(basically none of the geometry parameterization stuff). The purpose of this approach is to avoid having to make significant changes to the core components of OAS. I don't see anyone needing the geometry parameterization data downstream of AeroPoint. If it's needed for what ever reason they can get it from the Geometry Group itself.

openaerostruct/geometry/geometry_multi_sec_mesh.py
```diff
@@ -0,0 +1,731 @@
+""" Utility for quickly generating a multi-section user specificed OAS mesh
```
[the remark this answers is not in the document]
sabakhshi: This is a completely different approach to mesh generation that is useful for generating meshes for multi-section wings and single section wings from the same code. I was actually hoping to replace the current one with this in a future PR.

tests/multisection_integration_tests/test_multi_sym_2_sec_constr.py
sabakhshi: It should be. They are fundamentally the same problem with a different way of handling the constraint. However, I think SLSQP has trouble with this constraint causing the optimum to differ by a small error 10^-5, not within the 10^-6 tolerance. SNOPT doesn't have this issue. I can either reduce the tolerance or just use different reference value.

kanekosh: I think you should unify these two tests (test_multi_sym_2_sec.py and test_multi_sym_2_sec_constr.py) into one file/one class that runs two tests (i.e. has two methods). By doing so, you can reduce code duplication and set the reference value in one place (as an instance variable), which will be referred by the two tests. That way, you can test an output against the reference value and against the other way of handling geometry continuity.

And if SLSQP is causing the inconsistency here, I think you should use SNOPT for this test instead.

kanekosh: It looks like you're setting different chord_cp values for these two tests. Unless it is intentional and needed, it'd be better to set the same values for both and make sure that both method gives identical outputs.

sabakhshi: You're right about the different intial chord_cp values. I made them and the same but it still gives the same result. I think it's an optimizer issue since SLSQP does not handle this particular constraint well.

sabakhshi: Issue is taken care of.

openaerostruct/docs/advanced_features/scripts/basic_2_sec.py
```diff
@@ -0,0 +1,237 @@
+# docs checkpoint 0
```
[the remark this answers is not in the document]
sabakhshi: It shouldn't but none of the examples in the examples folder have doc checkpoint references like this one does. I can put copies in there though.

openaerostruct/docs/advanced_features/scripts/basic_2_sec.py
```diff
@@ -0,0 +1,237 @@
...
+prob.model.add_design_var("surface.sec0.chord_cp", lower=0.1, upper=10.0, units=None)
+prob.model.add_design_var("surface.sec1.chord_cp", lower=0.1, upper=10.0, units=None)
+prob.model.add_design_var("alpha", lower=0.0, upper=10.0, units="deg")
+
+
+# Next, we add the C0 continuity constraint for this problem by constraining the x-distance between sections to 0.
+# NOTE: SLSQP optimizer does not handle this equality constraint properly so the constraint needs to be specified as an inequality constraint
+# All other optimizers like SNOPT can handle the equality constraint as is.
+
+prob.model.add_constraint("surface.surface_joining.section_separation", upper=0, lower=0)  # FOR SLSQP
```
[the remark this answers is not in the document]
sabakhshi: I would prefer to leave it as is. It seems like SLSQP sometimes requires the scaling on this constraint to be changed in order to converge properly or may require that it be formulated as an inequality constraint. Since I'm not sure what optimizer the user will be using I prefer to leave it this way.

openaerostruct/utils/check_surface_dict.py
```diff
@@ -76,3 +76,90 @@ def check_surface_dict_keys(surface):
                 category=RuntimeWarning,
                 stacklevel=2,
             )
+
+
+def check_multi_sec_surface_dict_keys(surface):
```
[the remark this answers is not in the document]
sabakhshi: Done

sabakhshi:
Ready for review

openaerostruct/docs/advanced_features/scripts/basic_2_sec.py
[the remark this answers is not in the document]
kanekosh: Ok, I don't mind having doc checkpoints in the example file or keeping these in the current folder under docs as long as users can easily find them. And since you embedded all three files in the rst files it's good.
Either way, we should avoid putting two copies in difference places because that will make future maintenance difficult.

openaerostruct/utils/testing.py
```diff
@@ -217,6 +217,281 @@ def get_ground_effect_surfaces():
     return surfaces
 
 
+def get_three_section_surface_sym():
```
kanekosh, line 220: Could you clean-up these util functions and reduce code duplication?
For example, the only differences (I believe) between `get_three_section_surface_sym` and `get_three_section_surface_asym` are the `symmetry` and `root_section` entries. If so, you don't need to define `surface_dict` from scratch twice. You can first define it for let's say symmtry=True, and for the other util function, you can get that surface and overwrite `symmtry=False`.
Same for `get_two_section_surface_*`

tests/integration_tests/test_multi_2_sec.py
```diff
@@ -0,0 +1,409 @@
+from openmdao.utils.assert_utils import assert_near_equal
+import unittest
+
+
+class Test(unittest.TestCase):
+    def assert_check_results_sym(self, prob):
```
kanekosh, line 6: Could you add a quick comment for future reference, something like "we put assert_near_equal here and both test_constraints and test_constructions call this check because these two test cases should give the identical results"

tests/integration_tests/test_multi_2_sec.py
```diff
@@ -0,0 +1,409 @@
...
+        prob.driver.options["tol"] = 1e-3
+        prob.driver.options["disp"] = True
+        prob.driver.options["maxiter"] = 1000
+        # prob.driver.options["debug_print"] = ["nl_cons", "objs", "desvars"]
+
+        # Set up and run the optimization problem
+        prob.setup()
+        prob.run_driver()
+
+        assert_near_equal(prob["aero_point_0.surface_perf.CD"][0], 0.02849356, 1e-6)
```
kanekosh, line 298: I overlooked in the first review, but do you know why you get different results between the asym and sym cases? The surface dicts look like the same between two, so in theory, we should get the same results?

Is this because the b-spline works differently between sym and asym? If I understand it right, for a single-section surface, the root chord length = b-spline cp value at root for symmetry=True (because that's the anchor point), but for assymetric wing it's not always the case because the root chord length is not the anchor point but rather the "interpolated" value from b-spline. I don't know how this extends to the multi-section surface.

kanekosh (changes requested):

tests/integration_tests/test_multi_2_sec.py
sabakhshi: The asymmetrical and symmetrical 2 section and 3 sections cases are fundamentally different problems.  For example an asymmetrical 3 section case only features 3 section along the entire span while the symmetrical 3 section case mean that there there are 3 sections in the half span and therefore 6 sections total. These a fundamentally different problems and give different results.

tests/integration_tests/test_multi_2_sec.py
sabakhshi: Sure

openaerostruct/utils/testing.py
sabakhshi: Done

sabakhshi:
Ready for review

kanekosh (approved):

kanekosh (approved):

PR #449, opened 2024-12-11
Fix bugs and add moment output to moment coefficient component

sabakhshi:
## Purpose
Adds moment as a directly output from OAS in addition to CM. Fixes bugs related to derivative computation for CM. Adds significant commentary to the derivative code to explain what is going on. Cleans up code.

## Expected time until merged
1 week

## Type of change

- [x] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Some of the derivative updates need to be checked.

## Checklist

### DIFF
--- a/openaerostruct/functionals/moment_coefficient.py
+++ b/openaerostruct/functionals/moment_coefficient.py
@@ -5,7 +5,7 @@ import openmdao.api as om
 
 class MomentCoefficient(om.ExplicitComponent):
     """
-    Compute the coefficient of moment (CM) for the entire aircraft.
+    Compute the coefficient of moment (CM) and moment(M) for the entire aircraft.
 
     Parameters
     ----------
@@ -34,6 +34,8 @@ class MomentCoefficient(om.ExplicitComponent):
 
     Returns
     -------
+    M[3] : numpy array
+        The moment around the x-, y-, and z-axes at the cg point.
     CM[3] : numpy array
         The coefficient of moment around the x-, y-, and z-axes at the cg point.
     """
@@ -59,6 +61,7 @@ class MomentCoefficient(om.ExplicitComponent):
         self.add_input("S_ref_total", val=1.0, units="m**2", tags=["mphys_input"])
 
         self.add_output("CM", val=np.ones((3)), tags=["mphys_result"])
+        self.add_output("M", val=np.ones((3)), units="N*m", tags=["mphys_result"])
 
         self.declare_partials(of="*", wrt="*")
 
@@ -112,12 +115,15 @@ class MomentCoefficient(om.ExplicitComponent):
             # normalize CM
             if j == 0:
                 self.MAC_wing = MAC
+                self.S_ref_wing = S_ref
 
         self.M = M
 
+        # Output the moment vector
+        outputs["M"] = M
+
         # Compute the normalized CM
-        rho = inputs["rho"]
-        outputs["CM"] = M / (0.5 * rho * inputs["v"] ** 2 * inputs["S_ref_total"] * self.MAC_wing)
+        outputs["CM"] = M / (0.5 * inputs["rho"] * inputs["v"] ** 2 * inputs["S_ref_total"] * self.MAC_wing)
 
     def compute_partials(self, inputs, partials):
         cg = inputs["cg"]
@@ -128,12 +134,14 @@ class MomentCoefficient(om.ExplicitComponent):
         # Cached values
         M = self.M
         MAC_wing = self.MAC_wing
+        S_ref_wing = self.S_ref_wing
 
+        # Scaling factor of one over the dynamic pressure times sum of reference areas times the wing MAC
         fact = 1.0 / (0.5 * rho * v**2 * S_ref_total * MAC_wing)
 
-        partials["CM", "rho"] = -M * fact**2 * 0.5 * v**2 * S_ref_total * MAC_wing
-        partials["CM", "v"] = -M * fact**2 * rho * v * S_ref_total * MAC_wing
-        partials["CM", "S_ref_total"] = -M * fact**2 * 0.5 * rho * v**2 * MAC_wing
+        partials["CM", "rho"] = -M * fact / rho
+        partials["CM", "v"] = -2 * M * fact / v
+        partials["CM", "S_ref_total"] = -M * fact / S_ref_total
 
         partials["CM", "cg"][:] = 0.0
 
@@ -157,7 +165,7 @@ class MomentCoefficient(om.ExplicitComponent):
             MAC = 1.0 / S_ref * np.sum(panel_chords**2 * widths)
 
             # This produces a bi-diagonal matrix for the derivative of panel_chords with respect to chords
-            # This transformation matrix is further used for multiple derivatives
+            # This transformation matrix is further used for multiple derivatives later
             dpc_dc = np.zeros((ny - 1, ny))
             idx = np.arange(ny - 1)
             dpc_dc[idx, idx] = 0.5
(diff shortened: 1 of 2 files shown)


openaerostruct/functionals/moment_coefficient.py
```diff
@@ -112,12 +115,15 @@ def compute(self, inputs, outputs):
             # normalize CM
             if j == 0:
```
eytanadler, line 116: This isn't something you changed, but does this always assume that the main lifting surface (usually the wing) is the first one defined? If so, will the MAC, and thus the moment coefficient be wrong if the surfaces are defined in the wrong order? @kanekosh do you have any experience using multiple lifting surfaces? This seems error prone, but I'm not sure the best way to fix it...

openaerostruct/functionals/moment_coefficient.py
```diff
@@ -128,12 +134,14 @@ def compute_partials(self, inputs, partials):
...
 
+        # Scaling factor of one over the dynamic pressure times sum of reference areas times the wing MAC
         fact = 1.0 / (0.5 * rho * v**2 * S_ref_total * MAC_wing)
 
-        partials["CM", "rho"] = -M * fact**2 * 0.5 * v**2 * S_ref_total * MAC_wing
-        partials["CM", "v"] = -M * fact**2 * rho * v * S_ref_total * MAC_wing
-        partials["CM", "S_ref_total"] = -M * fact**2 * 0.5 * rho * v**2 * MAC_wing
+        partials["CM", "rho"] = -M * fact / rho
+        partials["CM", "v"] = -2 * M * fact / v
+        partials["CM", "S_ref_total"] = -M * fact / S_ref_total
```
eytanadler, line 144: Nice simplifications!

openaerostruct/functionals/moment_coefficient.py
```diff
@@ -174,34 +183,73 @@ def compute_partials(self, inputs, partials):
...
+            # distribution as the mid-panel chord distribution does wrt the the chord distribution.
             dc_dchord = np.einsum("ijkl,km->ijml", dcda, dpc_dc)
+
+            # Compute the derivative of CM wrt to the bound vortex points(b_pts) by reshaping dc_dchord to three rows
+            # and multiplying by fact.
             partials["CM", name + "_b_pts"] += dc_dchord.reshape((3, 3 * (nx - 1) * ny)) * fact
 
+            # Compute the derivative of M wrt to the bound vortex points(b_pts) by reshaping dc_dchord to three rows
+            # and multiplying by fact.
+            partials["M", name + "_b_pts"] += dc_dchord.reshape((3, 3 * (nx - 1) * ny))
```
eytanadler, line 248: I don't see a multiplication by `fact` as you mention in the comments

openaerostruct/functionals/moment_coefficient.py
```diff
@@ -216,28 +264,62 @@ def compute_partials(self, inputs, partials):
...
+                partials["M", name + "_b_pts"][2, :] = 0.0
                 dcda[0, :] = 0.0
                 dcda[1, :] *= 2.0
                 dcda[2, :] = 0.0
 
+            # Compute the derivative of CM wrt to the cg position which is negative dcda since diff = pts - cg times fact
+            # Accumlate the derivative over each surface as the total moment vector is sum over all surfaces.
             partials["CM", "cg"] -= dcda * fact
 
+            # Compute the derivative of M wrt to the cg position which is negative dcda since diff = pts - cg times fact
```
eytanadler, line 281: I think `fact` shouldn't be there at the end of the comment? Might be a copy/paste error or maybe I'm wrong

openaerostruct/functionals/moment_coefficient.py
```diff
@@ -216,28 +264,62 @@ def compute_partials(self, inputs, partials):
...
+                # base_dMAC_dS = dMAC_dS
             else:
                 # Apply this surface's portion of the moment to the MAC_wing term.
+                # We need to do this because CM is normalized by the MAC of the main wing
                 term = fact / (MAC_wing * MAC)
                 partials["CM", base_name + "_chords"] -= np.outer(M_j * term, base_dMAC_dc)
                 partials["CM", base_name + "_widths"] -= np.outer(M_j * term, base_dMAC_dw)
-                partials["CM", base_name + "_S_ref"] -= np.outer(M_j, base_dMAC_dS * term)
+                # partials["CM", base_name + "_S_ref"] -= np.outer(M_j, base_dMAC_dS * term)
+                partials["CM", base_name + "_S_ref"] = np.outer(M_j * fact, (-1/S_ref_total + 1/S_ref_wing))
```
eytanadler, line 325: This change and also why `term` has `MAC` and `MAC_wing` in the denominator in this `else` branch isn't totally clear to me. Could you explain the change in more detail?

tests/functionals_tests/test_moment_coefficient.py
```diff
@@ -16,7 +16,7 @@ def test(self):
 
         comp = MomentCoefficient(surfaces=surfaces)
 
-        run_test(self, comp)
+        run_test(self, comp, complex_flag=True, method='cs')
```
eytanadler, line 19: It sounds like you found a derivative error. Were these tests failing before? Or were they passing with bugs and not caught because they were using finite difference instead of complex step? Can you verify that they were failing without your changes if you switched to complex step?

eytanadler (commented):
Thanks @sabakhshi! More comments are always appreciated, particularly in complex derivative routines.

openaerostruct/functionals/moment_coefficient.py
sabakhshi: Yes. It assumes the first surface you define is the main wing. It's always been that way.

openaerostruct/functionals/moment_coefficient.py
sabakhshi: Error in the comment. fact is just the (1/q*S*MAC) term which normalizes to CM so we get rid of it when we do the moment derivative.

openaerostruct/functionals/moment_coefficient.py
sabakhshi: I actually don't think that is correct and I just fixed it

tests/functionals_tests/test_moment_coefficient.py
sabakhshi: Tests pass with fd too. I was just using cs while debugging and left it that way.

openaerostruct/functionals/moment_coefficient.py
kanekosh: Good catch... @sabakhshi how did you (or will you) deal with this for multisection surfaces? I think MAC of the 1st section doesn't make any sense if one physical surface is composed of multiple OAS "surfaces"?
Maybe we should move MAC computation into a separate component (and do some if-else there) and pass in the MAC as an input to this component?

kanekosh (changes requested):
Nice catch! I know you're working on it but I'm writing down as a note:
Test setup should be updated to be physically valid and to avoid false positive, which the current test does.

openaerostruct/functionals/moment_coefficient.py
sabakhshi: There is no need to worry about the multisection surfaces as they have been unified into a single surface before entering aeropoint.

kanekosh: @sabakhshi Would you mind documenting this (moment coefficient is always normalized by the 1st surface's MAC), maybe in advanced_features/Multiple Lifting Surfaces? One sentence noting this down would be enough.

tests/functionals_tests/test_moment_coefficient.py
```diff
@@ -36,7 +34,25 @@ def test2(self):
 
         group.connect("indep_var_comp.S_ref_total", "moment_calc.S_ref_total")
 
-        run_test(self, group)
+        run_test(self, group, complex_flag=True, method="cs")
+
+    def test3(self):
```
kanekosh, line 39: I don't have a strong opinion, but is there any reason you setup `test3` separately from `test2`? Maybe these two can be merged into one?

kanekosh (changes requested):

tests/functionals_tests/test_moment_coefficient.py
sabakhshi: Done

kanekosh (dismissed):

openaerostruct/functionals/moment_coefficient.py
```diff
@@ -5,7 +5,7 @@
 
 class MomentCoefficient(om.ExplicitComponent):
     """
-    Compute the coefficient of moment (CM) for the entire aircraft.
+    Compute the coefficient of moment (CM) and moment(M) for the entire aircraft.
```
eytanadler, line 8: Missing a space after "moment"

eytanadler (changes requested):
One more very minor comment. Could you also update the patch version for now so that we can track which versions have this fix while we wait on the numpy and openmdao update in #450?

openaerostruct/functionals/moment_coefficient.py
sabakhshi: Done!

eytanadler (approved):

PR #454, opened 2025-02-18
Added mesh shifting in multi-section unification component for maintaining LE coincidence

sabakhshi:
## Purpose
Added mesh shifting in multi-section unification component to allow for span/sweep transformations without a constraint. Previously, using a sectional scalar transformation was incompatible with the construction based approach. 
With this change the mesh is shifted inboard/outboard to compensate for span, sweep, and dihedral changes of subsequent sections.
The feature is optional but is on by default.
It is fully differentiated with analytic partials.

## Expected time until merged
3 days

## Type of change

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
New tests added/updated. Partials not complex safe so checks with FD enforced.

## Checklist


openaerostruct/docs/advanced_features/multi_section_surfaces.rst
```diff
@@ -39,32 +39,22 @@ For example, the tip chord or twist of a given section should match the root cho
...
-.. literalinclude:: /advanced_features/scripts/basic_2_sec_construction.py
-   :start-after: checkpoint 0
-   :end-before: checkpoint 1
-
-.. literalinclude:: /advanced_features/scripts/basic_2_sec_construction.py
-   :start-after: checkpoint 2
-   :end-before: checkpoint 3
+The number of linear constraints can quickly grow for problems with many sections.
+By reducing the degree of customization in section attachment, it is possible to eliminate these additional constraints and maintain C0 continuity using the construction-based approach.
+The constaint-based appraoch is described in a seperate section.
```
kanekosh, line 55: typo "approach"

openaerostruct/docs/advanced_features/multi_section_surfaces.rst
```diff
@@ -98,10 +87,38 @@ We then finish by plotting the result.
...
-
-
-
 The following shows the resulting optimized mesh. 
 The result is identical regardless of if the constraint-based or construction-based joining approahces are used.
 
 .. image:: /advanced_features/figs/multi_section_2_sym.png
+
+
+Construction-based Approach
```
kanekosh, line 96: Now that the construction-based approach is the default, would it make more sense to explain this first before talking about the constraint-based approach? Also "construction-based" may be not intuitive. Maybe we can call it just "default approach" vs. "(non-default) constraint-based approach" or something like that?

tests/multisection_tests/test_unification_comp.py
```diff
@@ -6,14 +6,22 @@
...
-        comp = GeomMultiUnification(sections=sec_dicts, surface_name=surface["name"])
+        comp = GeomMultiUnification(sections=sec_dicts, surface_name=surface["name"], shift_uni_mesh=False)
 
         run_test(self, comp, complex_flag=True, method="cs")
 
+    def test_shift(self):
+        (surface, chord_bspline) = get_three_section_surface()
+        sec_dicts = build_sections(surface)
+
+        comp = GeomMultiUnification(sections=sec_dicts, surface_name=surface["name"], shift_uni_mesh=True)
```
kanekosh, line 21: I think we should set a realistic mesh vaelus rather than all ones. You can do so either by manually setting mesh values or by including span/sweep/dihedral components.

kanekosh (changes requested):

tests/multisection_tests/test_unification_comp.py
sabakhshi: The get_three_section_surface() already retrieves a realistic 3 section surface for the test. I added a few a transformation in this test to make it more realistic.

openaerostruct/docs/advanced_features/multi_section_surfaces.rst
sabakhshi: Done

tests/multisection_tests/test_unification_comp.py
kanekosh: The surface dict has a realistic mesh but the OM variables do not because this test doesn't involve the component that set the mesh value from surface to OM problem. If you print the mesh `prob.get_val('comp.surface_uni_mesh', units='m')` you'll see all ones, which would miss the span/dihedral issue we talked about.
Can you add `set_val` or `set_inputs_default` to set the mesh values before run_test?

kanekosh (changes requested):

tests/multisection_tests/test_unification_comp.py
sabakhshi: Done

kanekosh: Span and sweep are good but the mesh doesn't have the dihedral (z=0 everywhere). I guess `build_sections` doesn't add dihedral, I don't know if that's intentional though. So please either fix `build_sections` or remove the `surface["dihedral"] = [-10, 10, -20]` line to avoid confusion

kanekosh (changes requested):

kanekosh (approved):

PR #468, opened 2025-10-06
Remove OM upper bound

kanekosh:
## Purpose

Remove OpenMDAO version upper bound now that OM 3.41 is released
## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

- [ ] I have run `ruff check` and `ruff format` to make sure the Python code adheres to PEP-8 and is consistently formatted


A-CGray:
Is the bug in question still present in OM 3.40? If so then we should exclude that version in the dependencies (I think you can do something like (`openmdao<3.40,>3.40`)

A-CGray (approved):

